PR #4, opened 2018-10-04
Calibration lists

rubyvanrooyen:
merging to master cleanup dirty updates in incorrect branch
and to get astokat-calc.py ready for proper development and use


astrokat/observatory.py
```diff
@@ -18,6 +18,33 @@ def __init__(self, location=None):
         self.mkat = self.get_location()
         self.observer = self.get_observer()
 
+    def _ephem_risetime_(self, ephem_target, lst=True):
```
wbode-za: class Observatory to inherit from object

PR #13, opened 2018-10-15
Obs script cleanup

rubyvanrooyen:
Version of script that is neater and verified to work on site.
Some updates and corrections remaining, but suggested that those be addressed after this branch is merged and specific devel branches can be created from master to address each issue

Conflict due to work on cadence observations will be handled by Ruby


astrokat/correlator.py
```diff
@@ -0,0 +1,26 @@
...
+        raise RuntimeError(msg)
+
+    for inp in session.cbf.fengine.inputs:
+        # Set the gain to a single non complex number if needed
+        if requant_gains is not None:
+            # TODO: read and store values before assignment
+            session.cbf.fengine.req.gain(inp, requant_gains)
+            msg = 'F-engine {} gain set to {}'.format(
+                    str(inp), requant_gains)
+            user_logger.info(msg)
```
lanceWilliams: You'll need to pass this function the logger reference.

astrokat/observatory.py
```diff
@@ -132,12 +114,15 @@ def collect_targets(kat, args):
             try:
                 catalogue.add(open(arg))
             except ValueError:
-                user_logger.warning('Catalogue {} contains bad targets'.format(arg))
+                msg = 'Catalogue {} contains bad targets'.format(arg)
+                user_logger.warningm(msg)
```
lanceWilliams: looks like a typo at 'warningm'

scripts/astrokat-observe.py
```diff
@@ -6,8 +6,7 @@
 import ephem
 
 import numpy as np
-from itertools import chain
-import time
+
 from datetime import datetime, timedelta
```
lanceWilliams: There are a couple of complaints on this file about indents and unused imports/variables. Maybe hit it with your linter again.

astrokat/correlator.py
rubyvanrooyen: This is only a function stub and not implemented.
After the merge I will branch of for the correlator setting implementation needed for beamformer.

I would like to leave it in a state that will cause and error, if that is okay

scripts/astrokat-observe.py
rubyvanrooyen: astrokat-observe.py has a number of conflicts, this is one of them, I am resolving this out now

astrokat/observatory.py
rubyvanrooyen: corrected, thank you

scripts/astrokat-observe.py
rubyvanrooyen: unused packages removed

rubyvanrooyen:
Remaining conflicts due to major changes in the scripts listed. This can only be resolved when merging with master

astrokat/noisediode.py
```diff
@@ -0,0 +1,67 @@
+import time
+
+from simulate import user_logger
+
+
+# switch noise-source pattern off
```
ajoubertza: Comment should be "on", not "off".  Also, this would be better as a docstring inside the function.

astrokat/noisediode.py
```diff
@@ -0,0 +1,67 @@
...
+def off(mkat, logging=True):
+    if logging:
+        user_logger.info('Switch noise-diode off')
+    # Noise Diodes are triggered on all antennas in array simultaneously
+    # add a second to ensure all digitisers set at the same time
+    timestamp = time.time() + 1
+    mkat.ants.req.dig_noise_source(timestamp, 0)
+
+
+# fire noisediode before track
```
ajoubertza: Please be consistent - some comments refer to "noise-source", some to "noisediode".
I like "noise diode".

astrokat/noisediode.py
```diff
@@ -0,0 +1,67 @@
...
+from simulate import user_logger
+
+
+# switch noise-source pattern off
+def on(mkat, logging=True):
+    if logging:
+        user_logger.info('Switch noise-diode on')
+    # Noise Diodes are triggered on all antennas in array simultaneously
+    # add a second to ensure all digitisers set at the same time
+    timestamp = time.time() + 1
```
ajoubertza: Rather change this `1` second to a constant at the start of the module - it is used in a few places.  E.g.
```
NOISE_DIODE_REQUEST_LEAD_TIME = 1
```

astrokat/noisediode.py
```diff
@@ -0,0 +1,67 @@
...
+        # add a second to ensure all digitisers set at the same time
+        timestamp = time.time() + 1
+        msg = 'Set all noise diode with timestamp {} ({})'.format(
+                int(timestamp), time.ctime(timestamp))
+        user_logger.info(msg)
+        mkat.ants.req.dig_noise_source(timestamp, on_fraction, cycle_length)
+    elif nd_pattern == 'cycle':
+        # Noise Diodes should trigger one after another, within reason
+        # add time [sec] to ensure all digitisers set at the same time
+        timestamp = time.time() + 2
```
ajoubertza: This lead time is 2 seconds, which differs from the others (1 sec).

astrokat/noisediode.py
```diff
@@ -0,0 +1,67 @@
...
+            ped.req.dig_noise_source(timestamp, on_fraction, cycle_length)
+            timestamp += cycle_length * on_fraction
+    elif nd_pattern in sub_ants:
+        # Noise Diodes are triggered for only one antenna in the array
+        ant_name = nd_pattern
+        user_logger.info('Set noise diode for antenna {}'.format(ant_name))
+        ped = getattr(mkat, ant_name)
+        ped.req.dig_noise_source('now', on_fraction, cycle_length)
+    else:
+        msg = 'Unknown ND cycle option, cannot apply requested pattern'
```
ajoubertza: Suggest including the value of `nd_pattern` in the error message to ease debugging.

PR #15, opened 2018-10-16
User/lance/add config to astrokat cals

lanceWilliams:
This PR is to add config reading to the astrokat-cals script if katconfig is available. First it will try the arg provided in --cat-path. If that is not a usable file, try the /var/kat/config path to find the file. If that path is not available, finally try to retrieve the config server uri from node conf and fetch the config that. 

I've added the bug label for now as the last case is not fully working. Testing on a node that can import katconf but does not have the /var/kat/config locally, the attempted import of the arrayconfig fails.

### DIFF
--- a/astrokat/observatory.py
+++ b/astrokat/observatory.py
@@ -1,4 +1,5 @@
 import os
+import json
 import ephem
 import numpy
 import katpoint
--- a/astrokat/observatory.py
+++ b/astrokat/observatory.py
@@ -8,24 +8,23 @@ from datetime import datetime
 from simulate import user_logger, setobserver
 from utility import katpoint_target, lst2utc
 
-
 try:
     import katconf
     # Set up configuration sourcei
-    config_path = '/var/kat/config'
-    node_file = '/var/kat/node.conf'
-    settings = {}
-    if os.path.isdir(config_path):
-        katconf.set_config(katconf.environ(override=config_path))
-    elif os.path.isfile(node_file):
-        with open(node_file, 'r') as fh:
-            node_conf = json.loads(fh.read())
-        for key, val in node_conf.items():
+    _config_path = '/var/kat/config'
+    _node_file = '/var/kat/node.conf'
+    _settings = {}
+    if os.path.isdir(_config_path):
+        katconf.set_config(katconf.environ(override=_config_path))
+    elif os.path.isfile(_node_file):
+        with open(_node_file, 'r') as fh:
+            _node_conf = json.loads(fh.read())
+        for key, val in _node_conf.items():
             # Remove comments at the end of the line
             val = val.split("#", 1)[0]
-            settings[key] = val.strip()
-            if node_conf.get("configuri", False):
-                katconf.set_config(katconf.environ(node_conf["configuri"]))
+            _settings[key] = val.strip()
+            if _node_conf.get("configuri", False):
+                katconf.set_config(katconf.environ(_node_conf["configuri"]))
             else:
                 raise ValueError("Could not open node config file using configuri")
     else:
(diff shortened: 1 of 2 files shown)


lanceWilliams:
Last use case is now working. Importing config from a node without the local files is possible. Requires a running CAM system to access the config server.

astrokat/observatory.py
```diff
@@ -6,9 +8,30 @@
 from simulate import user_logger, setobserver
 from utility import katpoint_target, lst2utc
 
+
 try:
     import katconf
-except ImportError:
+    # Set up configuration sourcei
```
ajoubertza: Typo?  `sourcei`

scripts/astrokat-cals.py
```diff
@@ -232,6 +226,34 @@ def get_cal(catalogue, target, ref_antenna):
 
 
 def main(args):
+
+    cam_config_path = '/var/kat/config'
+    node_file = '/var/kat/node.conf'
+    settings = {}
+
+    try:
+        import katconf
```
ajoubertza: All this import code is duplicated - would be nice if we could just have it once, in `observatory.py`.  Maybe something like
`from observatory import katconf`, although that could still raise an exception.  Alternatively, make a function for reading katconfig files in `observatory.py`, and use it around line 307.

astrokat/observatory.py
```diff
@@ -6,9 +8,30 @@
...
+    config_path = '/var/kat/config'
+    node_file = '/var/kat/node.conf'
+    settings = {}
```
ajoubertza: All the variables used in this code look like temporary ones, but they will appear in the `observatory` namespace, which could be confusing.  Could prefix with `_`, but probably better to put all this in a private function.

astrokat/observatory.py
```diff
@@ -53,6 +79,14 @@ def _ephem_settime_(self, ephem_target, lst=True):
         self.observer.date = set_time
         return self.observer.sidereal_time()
 
+    def _read_file_from_node_config(self, catalogue_file):
```
ajoubertza: This function is for public use, so no need for the underscore prefix.

astrokat/observatory.py
```diff
@@ -8,19 +10,43 @@
...
+    # Set up configuration source
+    _config_path = '/var/kat/config'
+    _node_file = '/var/kat/node.conf'
+    _settings = {}
+    if os.path.isdir(_config_path):
+        katconf.set_config(katconf.environ(override=_config_path))
+    elif os.path.isfile(_node_file):
+        with open(_node_file, 'r') as fh:
+            _node_conf = json.loads(fh.read())
+        for key, val in _node_conf.items():
```
ajoubertza: Won't the `key`, `value` and `val` variables also end up in the namespace?

astrokat/observatory.py
```diff
@@ -8,19 +10,43 @@
...
+                katconf.set_config(katconf.environ(_node_conf["configuri"]))
+            else:
+                raise ValueError("Could not open node config file using configuri")
+    else:
+        raise ValueError("Could not open node config file")
+
+except (ImportError, ValueError):
     # default reference position for MKAT array
-    ref_location = 'ref, -30:42:47.4, 21:26:38.0, 1060.0, 0.0, , , 1.15'
+    _ref_location = 'ref, -30:42:47.4, 21:26:38.0, 1060.0, 0.0, , , 1.15'
```
ajoubertza: Please update to match the current location:  https://github.com/ska-sa/katconfig/blob/karoo/static/arrays/karoo.array_mkat.conf#L21

astrokat/observatory.py
```diff
@@ -8,19 +10,43 @@
...
+    if os.path.isdir(_config_path):
+        katconf.set_config(katconf.environ(override=_config_path))
+    elif os.path.isfile(_node_file):
+        with open(_node_file, 'r') as fh:
+            _node_conf = json.loads(fh.read())
+        for _key, _val in _node_conf.items():
+            # Remove comments at the end of the line
+            _val = _val.split("#", 1)[0]
+            _settings[_key] = _val.strip()
+            if _node_conf.get("configuri", False):
```
ajoubertza: Please move this outside the loop and use `_settings`, not `_node_conf`.

PR #19, opened 2018-10-23
Calvin users

rubyvanrooyen:
User input updates mainly, not changing much of the base functionality, but adding more "niceness"


rubyvanrooyen:
Looking at this pull request again, I think it is premature
I am going to spin the noise diode development from this branch and this merge here, from where a full verification can be done
Closing this pull request, will issue a new one when noise diode implementation is included

PR #34, opened 2019-01-15
Add `katconf`, `katcorelib` as extra dependencies

sw00:
These additional packages are used in the live system but were not specified as dependencies.

I'm adding them as optional extras (can be installed using `pip install -e .[live]`). See [PEP508](https://www.python.org/dev/peps/pep-0508/) for more information.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,4 +7,7 @@ setup(
                       'katpoint',
                       'matplotlib',
                       'numpy',
-                      'pyyaml'])
+                      'pyyaml'],
+    extras_require={
+        'live': ['katcorelib', 'katconf']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -7,4 +7,7 @@
                       'katpoint',
                       'matplotlib',
                       'numpy',
-                      'pyyaml'])
+                      'pyyaml'],
+    extras_require={
+        'live': ['katcorelib', 'katconf']
```
ajoubertza: This is a nice solution.
(Merging will likely create a conflict on PR #33, but we'll deal with that there)

Issue #14, opened 2018-10-16
setup.py missing to pip install astrokat


wbode-za:
A setup.py file is needed to allow CAM to install astrokat as a package.
What are all of the astrokat dependencies?

SKAJohanVenter:
@wbode-za Only saw this issue now. I added a bare bones `setup.py` in PR  https://github.com/ska-sa/astrokat/pull/16 that can be used as is, but should really be expanded.

rubyvanrooyen:
astrokat package now pip installable (karoocamv22 release)

PR #39, opened 2019-01-29
Add katversion support and install scripts

ajoubertza:
- Add _katversion_ so that we can get version strings derived from the git version info.  This is required for CAM's automated branch creation tool.
- Install the useful scripts so that they are available on the path.
- Add some more details to setup.py.

Closes issue #14

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 
 # /
+.eggs
 *.egg-info
 *.pyc
 *~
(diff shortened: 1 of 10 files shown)


rubyvanrooyen:
@ajoubertza I have versions hardcoded in the scripts, should I go through and remove those?

ajoubertza:
@rubyvanrooyen Oh, I see that now.  Yes, I can remove them, or make them report the astrokat version instead.  Which do you prefer?

rubyvanrooyen:
@ajoubertza can we remove the hardcoded values please

ajoubertza:
@rubyvanrooyen Can you have a look at this again, please?

PR #46, opened 2019-02-11
Verify and test

rubyvanrooyen:
Need to get jenkins tests to past, but updates now with integration test scripts and obs_plans

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -391,30 +391,29 @@ def run_observation(opts, kat):
                             .format(observer))
 
             # Verify the observation is in a valid LST range
-            if not kat.array.dry_run:
-                # Verify that it is worth while continuing with the observation
-                # Do not use float() values, ephem.hours does not convert as expected
-                local_lst = observer.sidereal_time()
-                # Only observe targets in current LST range
-                if start_lst < end_lst:
-                    in_range = ((ephem.hours(local_lst) > ephem.hours(str(start_lst)))
-                                and (ephem.hours(local_lst) < ephem.hours(str(end_lst))))
-                    if not in_range:
-                        user_logger.error('Local LST {} outside LST range {}-{}'
-                                          .format(local_lst,
-                                                  start_lst,
-                                                  end_lst))
-                        continue
-                else:
-                    # else assume rollover at midnight to next day
-                    out_range = ((ephem.hours(local_lst) < ephem.hours(str(start_lst)))
-                                 and (ephem.hours(local_lst) > ephem.hours(str(end_lst))))
-                    if out_range:
-                        user_logger.warning('Local LST {} outside LST range {}-{}'
-                                            .format(local_lst,
-                                                    start_lst,
-                                                    end_lst))
-                        continue
+            # and that it is worth while continuing with the observation
+            # Do not use float() values, ephem.hours does not convert as expected
+            local_lst = observer.sidereal_time()
+            # Only observe targets in current LST range
+            if start_lst < end_lst:
+                in_range = ((ephem.hours(local_lst) > ephem.hours(str(start_lst)))
+                            and (ephem.hours(local_lst) < ephem.hours(str(end_lst))))
+                if not in_range:
+                    user_logger.error('Local LST {} outside LST range {}-{}'
+                                      .format(local_lst,
+                                              start_lst,
+                                              end_lst))
+                    continue
+            else:
+                # else assume rollover at midnight to next day
+                out_range = ((ephem.hours(local_lst) < ephem.hours(str(start_lst)))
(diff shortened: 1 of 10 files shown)


astrokat/observe_main.py
```diff
@@ -477,7 +475,8 @@ def run_observation(opts, kat):
                         if not tgt:
                             break
                         # check enough time remaining to continue
-                        if time_remaining < tgt['duration']:
+                        if obs_duration > 0 and\
+                           time_remaining < tgt['duration']:
```
ajoubertza: Rather use parenthesis, as per our coding guidelines:
```python
if (obs_duration > 0 and
        time_remaining < tgt['duration']):
    done = True
    break
```
Or you could do it all on one line - that would fit within our 90 char limit.

astrokat/simulate.py
```diff
@@ -135,6 +136,7 @@ def __iter__(self):
         raise StopIteration
 
     def __exit__(self, type, value, traceback):
+        # TODO: what is self.track_ used for again?
```
ajoubertza: This is a bit concerning - would prefer all traces are removed, if there is no purpose.

astrokat/simulate.py
```diff
@@ -135,6 +136,7 @@ def __iter__(self):
...
         if self.track_:
```
ajoubertza: No longer creating this variable in `__init__`, so can we be sure it will even exist?

astrokat/utility.py
```diff
@@ -17,6 +17,45 @@ class NoTargetsUpError(Exception):
...
+        if data['durations'] is None:
+            msg = 'durations primary key cannot be empty in observation YAML file'
+            raise RuntimeError(msg)
+        if 'start_time' in data['durations']:
+            start_time = data['durations']['start_time']
+            if type(start_time) is str:
+                data['durations']['start_time'] = \
+                        datetime.datetime.strptime(start_time, '%Y-%m-%d %H:%M')
+    if 'observation_loop' not in data.keys():
+        raise RuntimeError('Nothing to observer, exiting')
```
ajoubertza: Typo:  observer -> observe

astrokat/utility.py
```diff
@@ -17,6 +17,45 @@ class NoTargetsUpError(Exception):
...
+    if data['observation_loop'] is None:
+        raise RuntimeError('Empty observation loop, exiting')
+    for obs_loop in data['observation_loop']:
+        if type(obs_loop) is str:
+            raise RuntimeError('Expected observation list, got string')
```
ajoubertza: Error says a list is expected, but the next line uses `obs_loop.keys()`, which suggest it is more like a dict?

astrokat/utility.py
```diff
@@ -17,6 +17,45 @@ class NoTargetsUpError(Exception):
 def read_yaml(filename):
     with open(filename, 'r') as stream:
         data = yaml.safe_load(stream)
+
+    # handle mapping of user friendly keys to CAM resource keys
```
ajoubertza: Better that this is in utils now - thanks.

astrokat/observe_main.py
rubyvanrooyen: Added the parenthesis implementation , tx

astrokat/simulate.py
rubyvanrooyen: This is for multiple observation loops, needed for zoom modes
I never got time to finish the implementation, sort of like the correlator setup implementation

These are the 2 big hanging implementations that I still need to do and will add separate branches to fix them

astrokat/simulate.py
rubyvanrooyen: Incomplete implementation, only realised the inconsistency when I wrote the planning integration scripts

This must be completed for multiple observation loops, will be addressed in separate devel branch

astrokat/utility.py
rubyvanrooyen: Yes, the YAML is read into a JSON style dictionary, but this is for the user input.
In the YAML the observation loops are specified as having an LST range + a list of targets.
Thus, giving an error that is not "quite" correct makes the required input a little clearer.

What about extending the error message to something like?
"
RuntimeError("Incomplete observation input: LST range and list of observations required.")
"

rubyvanrooyen: Suggested implementation
RuntimeError("Incomplete observation input: LST range and a minimum of one target required.")

rubyvanrooyen: Updated error message implemented for more explicit information to user related to possible error.
+
Added TODO note that needs to be implemented and verified with the implementation of multiple observation loop needed for zoom modes.
Issue #47

rubyvanrooyen:
All requested changes have been applied.
Merging to master to unify code

PR #51, opened 2019-04-05
Port Jenkinsfile to Declarative Pipeline Style

02Anele:
Port to new declarative syntax for Jenkins pipeline definitions in an attempting to fix our broken multibranch jobs because our new Jenkins agents inside containers don't have access to Docker daemon. this pull request is inline with a JIRA: [ MT-610](https://skaafrica.atlassian.net/secure/RapidBoard.jspa?rapidView=40&projectKey=MT&modal=detail&selectedIssue=MT-610), for reference please check this PR: [141](https://github.com/ska-sa/katconf/pull/141)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -43,9 +43,9 @@ pipeline {
         }
 
         stage('Archive build artifact: .whl & .deb'){
-             steps {
+            steps {
                 archiveArtifacts 'dist/*'
-             }
+            }
         }
 
         stage('Trigger downstream publish') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,14 +5,16 @@ pipeline {
 
     stages {
         stage('Checkout SCM') {
-            checkout([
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,43 +1,61 @@
...
-                timeout(time: 30, unit: 'MINUTES') {
-                    try {
+        stage('Install & Unit Tests') {
+            options {
+                timestamps()
+                timeout(time: 30, unit: 'MINUTES') 
+            }
+            steps
+                {
                         sh 'pip install . -U --pre'
```
mmphego: LGTM

As a side note, perhaps we should move away from running tests with `python setup.py nosetests` instead use `tox`.
Look at https://github.com/ska-sa/katconf/blob/master/Jenkinsfile

But for now, looks good.

ajoubertza: @mmphego We will move to tox as we do the Python 2 to 3 migration.

PR #53, opened 2019-04-15
Trigger `ci.publish-artifacts` for multibranch job

sw00:
- Add static analysis stage
- Trigger downstream publish when on master branch

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,10 @@ pipeline {
         label 'cambuilder'
     }
 
+    environment {
+        KATPACKAGE = "${(env.JOB_NAME - env.JOB_BASE_NAME) - '-multibranch/'}"
+    }
+
     stages {
         stage('Checkout SCM') {
             steps{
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -17,6 +17,17 @@ pipeline {
             }
         }
 
+        stage ('Static analysis') {
+            steps {
+                sh 'pylint katcomp --output-format=parseable --exit-zero > pylint.out'
```
ajoubertza: `katcomp`?

sw00: The src directory. Probably could be more explicit written as `./katcomp`. Will test it out.

Jenkinsfile
```diff
@@ -52,28 +52,21 @@ pipeline {
                 }
             }
 
-        stage('Build .whl & .deb') {
+        stage('Build & publish packages') {
```
sw00: @ajoubertza I've decided to merge building packages and triggering their publish into one stage. There's no point building packages if they're not going to be published. This will save time on resources on all non-master branch builds.

Any issues you foresee with this?

ajoubertza: Is it possible for the tests to pass, but a build to fail?  Then we might want test builds before merging.  I don't recall ever seeing a build of one of our packages failing (other than katgui, but that's not Python).

sw00: True, any exception (non-zero exit) can cause a build to fail. However simply having the tests pass as the criteria for merging to `master` is a reasonable quality gate. 

The package step might still fail afterward (but this will still halt the release pipeline).

PR #54, opened 2019-04-26
Verification after merge

rubyvanrooyen:
Updates and correction identified after merge with master for V22 rollout


astrokat/__main__.py
```diff
@@ -16,17 +15,19 @@ def session_options(parser,
                     x_short_opts=[],
                     x_long_opts=[]):
```
ajoubertza: The meaning of the `x` prefix isn't clear to me.  After looking through the code it seems to mean "options that must be removed".  Suggest the names are made a little more obvious.  E.g.
`hidden_short_opts` and `hidden_long_opts`,
or
`short_opts_to_remove` and `long_opts_to_remove`.

astrokat/correlator.py
```diff
@@ -1,10 +1,12 @@
...
+
 def set_fengines(session,
-        requant_gains=None,
-        fft_shift=None):
+                 requant_gains=None,
+                 fft_shift=None):
+    msg = 'Failed to get correlator input labels: '
     if not session.cbf.fengine.inputs:
-        msg = 'Failed to get correlator input labels, cannot set the F-engine gains'
+        msg += 'cannot set the F-engine gains'
```
ajoubertza: This change seems unnecessary.  The base `msg` isn't used anywhere else - lines 17 and 24 reassign it.

astrokat/utility.py
```diff
@@ -133,7 +137,9 @@ def get_lst(yaml_lst):
         start_lst = time_.hour + time_.minute/60.
 
     if end_lst is None:
-        end_lst = (start_lst + 12.) % 24.
+        end_lst = (start_lst + 24.) % 24.
+        if numpy.abs(end_lst - start_lst) < 1.:
+            end_lst = 24.
```
ajoubertza: Why is this done?  If start and end LST are within 1 hour, then force end LST to midnight?
Looks like it will often be the case, e.g. start LST values from `[0..24)` would end up here.
Not obvious for a non-astronomer like me 🙂

astrokat/__main__.py
rubyvanrooyen: updated to short_opts_to_remove and long_opts_to_remove

astrokat/correlator.py
rubyvanrooyen: Cleaned up implementation to single message

astrokat/utility.py
rubyvanrooyen: It is to stabilise the math -- mod of 0 (zero) give zero, then end and start is both zero and nothing happens, even though you want it to run over 24 hours

Plus, I need to include 24 -- even though technically I never reach it -- I had some commissioning astronomers complain if I do not set this to explicit 24

ajoubertza: So if unspecified, end LST is always 24, unless start LST >= 24:
```python
In [1]: import numpy

In [2]: def calc_end_lst(start_lst):
   ...:     end_lst = (start_lst + 24.) % 24.
   ...:     if numpy.abs(end_lst - start_lst) < 1.:
   ...:         end_lst = 24.
   ...:     return end_lst
   ...: 

In [3]: for start_lst in [0., 1., 6., 12., 18., 24., 30.]:
   ...:     end_lst = calc_end_lst(start_lst)
   ...:     print('start_lst {} .. end_lst {}'.format(start_lst, end_lst))
   ...:     
start_lst 0.0 .. end_lst 24.0
start_lst 1.0 .. end_lst 24.0
start_lst 6.0 .. end_lst 24.0
start_lst 12.0 .. end_lst 24.0
start_lst 18.0 .. end_lst 24.0
start_lst 24.0 .. end_lst 0.0
start_lst 30.0 .. end_lst 6.0
```

astrokat/observe_main.py
```diff
@@ -313,6 +319,7 @@ def run_observation(opts, kat):
     obs_plan_params = opts.obs_plan_params
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
+    local_horizon = obs_plan_params['instrument']['horizon']
```
ajoubertza: Is the value also going to end up in CAM's session object?  E.g. with the `--horizon` option.  If not, the value in the yaml file could differ from session's meaning things like `session.track` fail or succeed unexpectedly.

ajoubertza:
We also need the `Jenkinsfile` from the master branch copied into this branch so that the tests can run again.

astrokat/utility.py
rubyvanrooyen: Yes, exactly, LST = 0 - 24 floating hours (decimal).
Normally a single target LST range is 8 to 10 hours, visible above the horizon, which is 4 minutes earlier each day (celestial time), thus %24 

But zero (0) is never our friend, so special checks need to be implemented if you have a mod value that is a factor of 24, but close to the a full 24 hour time difference.

When will this happen?
Normally if we have targets that has been chosen to always have at least one target visible no matter what time in a 24 hour cycle the observation get scheduled.

So in short, one of those nasty times when astronomy may cause our math to fail in practise, but still be valid in theory....

astrokat/utility.py
```diff
@@ -16,35 +18,50 @@ class NoTargetsUpError(Exception):
 # Read config .yaml file
 def read_yaml(filename):
     with open(filename, 'r') as stream:
-        data = yaml.safe_load(stream)
-
+        try:
+            data = yaml.safe_load(stream)
+        except yaml.parser.ParserError:
+            return False
```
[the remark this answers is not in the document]
rubyvanrooyen: Error in programming, should return empty dict, corrected

astrokat/observe_main.py
```diff
@@ -626,18 +631,16 @@ def main(args):
...
-            # catch other hidden arguments such as correlator settings
-            if len(arg.split('=')[1]) > 1:
-                arg = "{}='{}'".format(arg.split('=')[0], arg.split('=')[1])
-            if arg.startswith(("-", "--")):
-                parser.add_argument(arg)
-        args_ = parser.parse_args(args)
+    # astrokat does not support observational command line parameters
+    # all observational parameters must be in the YAML file
+    # command line arguments will be dropped to unknown arguments
+    unknown_args = [arg for arg in args if '--' in arg]
```
ajoubertza: Rather `if arg.startswith('--')` than `if '--' in arg`.

astrokat/observe_main.py
```diff
@@ -626,18 +631,16 @@ def main(args):
...
+    if '--horizon' in unknown_args:
```
ajoubertza: The instruction sets often use the form `--horizon=20`, so this won't find it.  Could loop through all unknown and check if any start with `--horizon`.

astrokat/observe_main.py
```diff
@@ -340,17 +351,16 @@ def run_observation(opts, kat):
...
         [start_lst, end_lst] = get_lst(observation_cycle['LST'])
 
-
         # Verify that it is worth while continuing with the observation
         # The filter functions uses the current time as timestamps
         # and thus incorrectly set the simulation timestamp
         if not kat.array.dry_run:
             # Quit early if there are no sources to observe
-            if len(catalogue.filter(el_limit_deg=opts.horizon)) == 0:
+            if len(catalogue.filter(el_limit_deg=local_horizon)) == 0:
```
ajoubertza: We need the horizon value to be set in `opts.horizon`, since this is how it gets into session (when we call `session.standard_setup` on line 386).  Rather drop the `local_horizon` variable, and stick with `opts.horizon`.

astrokat/observe_main.py
```diff
@@ -314,6 +319,12 @@ def run_observation(opts, kat):
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
 
+    # set local horizon
+    if 'horizon' in obs_plan_params:
+        local_horizon = obs_plan_params['horizon']
+    else:
+        local_horizon = 20.  # deg above horizon default
```
ajoubertza: As noted in another comment, rather overwrite `opts.horizon` here and line 324.

astrokat/observe_main.py
rubyvanrooyen: done

rubyvanrooyen: if any('horizon' in arg for arg in unknown_args):

astrokat/observe_main.py
rubyvanrooyen: did the assignment in the wrong place -- corrected it to assign to opts.horizon to ensure sessions has the value and then only use the session value in the observation, as it should be
need to build a test to check this

astrokat/observe_main.py
rubyvanrooyen: yaml does not support --horizon anymore -- will get value and assign to session to ensure basic functionality

ajoubertza:
Commits 1d5e7d7 and 6ebb65b were not supposed to be added to this branch, so I've forced it back to 9542830.

ajoubertza:
Merging now.

PR #55, opened 2019-06-07
astrokat learning and rework 4.1

xinyuwu:
[MT-758](https://skaafrica.atlassian.net/browse/MT-758)

- rewrote the python unit test to check output
- made a couple of changes for python 3 compatibility
- changed integration_time to integration_period. Ruby has made the same changes

I will merge my code with Ruby's after this PR is approved.


ajoubertza:
@xinyuwu Please edit this PR and change the base to `master`.  That is where we want to merge to.

yaml_list.yaml
```diff
@@ -1,7 +1,14 @@
 yaml_test_list:
```
ajoubertza: The code to loop through the items in this file:
```python
for test in config['yaml_test_list']:
```
has been removed, so might as well remove the file too.

Similarly, the `nose-testconfig` requirements in `setup.py`, and the `tc-file` and `tc-format` options in `setup.cfg` won't be needed anymore.

astrokat/__init__.py
```diff
@@ -1,7 +1,11 @@
+from __future__ import division
```
wbode-za:
```suggestion
from __future__ import absolute_import, division
```

wbode-za: Applies to the others as well...

astrokat/__init__.py
```diff
@@ -1,7 +1,11 @@
...
+from __future__ import absolute_import
+
 from .__main__ import cli  # noqa
-import noisediode  # noqa
-import correlator  # noqa
-import scans # noqa
+
+from .noisediode  import * # noqa
+from .correlator  import * # noqa
+from .scans import * # noqa
```
ajoubertza: This doesn't look the same as it used to be.  Previously we imported the modules not everything inside them - this will change the global namespace, or at least pollute it.  E.g. If we've done `import astrokat`, then we can access the `noisediode`'s `on` function as `astrokat.on`.

Rather:
```python
from . import noisediode
from . import correlator
from . import scans
```

astrokat/noisediode.py
```diff
@@ -1,10 +1,15 @@
...
 import time
+
 libnames = ['user_logger']
+
 try:
-    lib = __import__('katcorelib', globals(), locals(), libnames, -1)
+    lib = __import__('katcorelib', globals(), locals(), libnames)
 except ImportError:
-    lib = __import__('simulate', globals(), locals(), libnames, -1)
+    lib = __import__('astrokat.simulate', globals(), locals(), libnames)
```
ajoubertza: Can't we drop the `astrokat` prefix here, so just `.simulate`?

(Also wonder if there is a better way to do these imports - it isn't pretty)

astrokat/observe_main.py
```diff
@@ -401,7 +401,7 @@ def run_observation(opts, kat):
             # Do not use float() values, ephem.hours does not convert as expected
             local_lst = observer.sidereal_time()
             # Only observe targets in current LST range
-            if start_lst < end_lst:
+            if float(start_lst) < end_lst:
```
ajoubertza: If `start_lst` might be a string, are we sure that `end_lst` is always a float?

astrokat/scans.py
```diff
@@ -1,11 +1,17 @@
...
+
 import time
+
 libnames = ['user_logger']
 try:
-    lib = __import__('katcorelib', globals(), locals(), libnames, -1)
+    lib = __import__('katcorelib', globals(), locals(), libnames)
 except ImportError:
-    lib = __import__('simulate', globals(), locals(), libnames, -1)
+    lib = __import__('astrokat.simulate', globals(), locals(), libnames)
```
ajoubertza: `.simulate`?

astrokat/observe_main.py
```diff
@@ -18,17 +18,10 @@
...
-            'verify_and_connect']
 try:
-    lib = __import__('katcorelib', globals(), locals(), libnames)
+    from katcorelib import collect_targets, user_logger, start_session, verify_and_connect
 except ImportError:
-    lib = __import__('astrokat', globals(), locals(), libnames)
-finally:
-    for libname in libnames:
-        globals()[libname] = getattr(lib, libname)
+    from astrokat import collect_targets, user_logger, start_session, verify_and_connect
```
ajoubertza: Does `from . import` work?  That would be better.

PR #58, opened 2019-07-15
pin numpy to versions with py2 support < 1.17

samtwum:
Numpy's 1.17 pre-release has dropped support to python 2. See numpy [repo](https://github.com/numpy/numpy/releases/tag/v1.17.0rc1).

This PR is to ensure that the numpy install < v1.17 for py2 and >=v1.17 for py3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
     install_requires=['pyephem',
                       'katpoint',
                       'matplotlib<3',
-                      'numpy<1.17',
+                      'numpy>=1.17' if sys.version_info >= (3, 5) else 'numpy<1.17',
                       'pyyaml'],
     extras_require={
         'live': ['katcorelib', 'katconf']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,7 +39,7 @@
     install_requires=['pyephem',
                       'katpoint',
                       'matplotlib<3',
-                      'numpy',
+                      'numpy<1.17',
```
ajoubertza: This is fine for Python 2, but astrokat can also run under Python 3, and no need to limit  it there.
We can do something like this, to have different packages on different versions:  https://github.com/ska-sa/katportalclient/blob/master/setup.py#L41-L42

setup.py
```diff
@@ -39,7 +40,7 @@
     install_requires=['pyephem',
                       'katpoint',
                       'matplotlib<3',
-                      'numpy',
+                      'numpy>=1.17' if sys.version_info >= (3, 5) else 'numpy<1.17',
```
ajoubertza: Tests are failing for Python 3 as 1.17 isn't available yet.  See comment in katconf PR.

samtwum:
Refer to https://github.com/ska-sa/katdeploy/pull/424

PR #59, opened 2019-07-16
Noisediode debugging

rubyvanrooyen:
Correct noise diode period calculation to be multiple of correlator dump period and < 20 sec for L-band
Log message return actual digitiser timestamps when available
Cleanup bad symbolic links

### DIFF
--- a/astrokat/noisediode.py
+++ b/astrokat/noisediode.py
@@ -16,7 +16,10 @@ _DEFAULT_LEAD_TIME = 5.0  # lead time [sec]
 def _katcp_reply_to_log_(dig_katcp_replies):
     for ant in sorted(dig_katcp_replies):
         reply, informs = dig_katcp_replies[ant]
-        _nd_log_msg_(ant, reply, informs)
+        # assuming ND for all antennas set at the same time
+        # only return timestamp of the last antenna set
+        timestamp = _nd_log_msg_(ant, reply, informs)
+    return timestamp
 
 
 def _nd_log_msg_(ant,
@@ -37,11 +40,13 @@ def _nd_log_msg_(ant,
     msg = ('Noise diode for antenna {} set at {}. '
            .format(ant,
                    actual_time))
-    user_logger.debug(msg)
+    user_logger.info(msg)
     msg = ('Pattern set as {} sec ON for {} sec cycle length'
            .format(actual_on_frac*actual_cycle,
                    actual_cycle))
-    user_logger.debug(msg)
+    user_logger.info(msg)
+
+    return actual_time
 
 
 # switch noise-source on
@@ -68,13 +73,21 @@ def on(kat,
     user_logger.trace('TRACE: nd on at {} ({})'
                       .format(timestamp,
                               time.ctime(timestamp)))
-    msg = ('Switch noise-diode on at {}'
+    msg = ('Request: Switch noise-diode on at {}'
            .format(timestamp))
     user_logger.info(msg)
 
     replies = kat.ants.req.dig_noise_source(timestamp, 1)
     if not kat.dry_run:
-        _katcp_reply_to_log_(replies)
+        timestamp = _katcp_reply_to_log_(replies)
+    else:
+        # - use integer second boundary as that is most likely be an exact
+        #   time that DMC can execute at
+        msg = ('Dry-run: Set all noise diodes with timestamp {} ({})'
+               .format(np.ceil(timestamp),
+                       time.ctime(timestamp)))
+        user_logger.info(msg)
+
     return timestamp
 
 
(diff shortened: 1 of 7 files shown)


astrokat/noisediode.py
```diff
@@ -68,13 +73,21 @@ def on(kat,
...
            .format(timestamp))
     user_logger.info(msg)
 
     replies = kat.ants.req.dig_noise_source(timestamp, 1)
     if not kat.dry_run:
-        _katcp_reply_to_log_(replies)
+        timestamp = _katcp_reply_to_log_(replies)
+    else:
+        # - use integer second boundary as that is most likely be an exact
+        #   time that DMC can execute at
```
ajoubertza: This comment is more appropriate around line 72 - that is where apply the `ceil` to the time.  And it should happen regardless of the dry-run state.

astrokat/noisediode.py
```diff
@@ -68,13 +73,21 @@ def on(kat,
...
+        msg = ('Dry-run: Set all noise diodes with timestamp {} ({})'
+               .format(np.ceil(timestamp),
```
ajoubertza: If the caller passed in a timestamp, if might have a fractional part, and we assume they know what they are doing.  We should log that time.  The DMC will try to get close to that time, even if it isn't on an integer boundary.

(Also note that the ctime calling isn't using the `ceil` of the timestamp, so it could differ slightly - we should be consistent)

astrokat/noisediode.py
```diff
@@ -104,15 +117,22 @@ def off(kat,
     user_logger.trace('TRACE: nd off at {} ({})'
                       .format(timestamp,
                               time.ctime(timestamp)))
-    msg = ('Switch noise-diode off at {}'
+    msg = ('Request: Switch noise-diode off at {}'
            .format(timestamp))
     user_logger.info(msg)
 
     # Noise Diodes are triggered on all antennas in array simultaneously
     # add a second to ensure all digitisers set at the same time
```
ajoubertza: These two comments can be removed - just a repeat of lines 110-111.

astrokat/noisediode.py
```diff
@@ -104,15 +117,22 @@ def off(kat,
...
     replies = kat.ants.req.dig_noise_source(timestamp, 0)
     if not kat.dry_run:
-        _katcp_reply_to_log_(replies)
+        timestamp = _katcp_reply_to_log_(replies)
+    else:
+        # - use integer second boundary as that is most likely be an exact
+        #   time that DMC can execute at
+        msg = ('Dry-run: Set all noise diodes with timestamp {} ({})'
+               .format(np.ceil(timestamp),
```
ajoubertza: Ditto on not using `ceil`, and moving comment to around line 116.

astrokat/noisediode.py
```diff
@@ -219,20 +240,22 @@ def pattern(kat,  # kat subarray object
...
-           'with {} sec lead time'
+    msg = ('Request: Set noise diode pattern to activate at {} '
+           '(includes {} sec lead time)'
            .format(timestamp,
                    lead_time))
     user_logger.warning(msg)
 
     if nd_antennas == 'all':
         # Noise Diodes are triggered on all antennas in array simultaneously
         # add a second to ensure all digitisers set at the same time
```
ajoubertza: This comment isn't necessary anymore - already noted on lines 240-241.

astrokat/noisediode.py
```diff
@@ -219,20 +240,22 @@ def pattern(kat,  # kat subarray object
...
+        replies = kat.ants.req.dig_noise_source(timestamp,
+                                                on_fraction,
+                                                cycle_length)
         if not kat.dry_run:
-            _katcp_reply_to_log_(replies)
+            timestamp = _katcp_reply_to_log_(replies)
         else:
-            msg = ('Set all noise diodes with timestamp {} ({})'
+            msg = ('Dry-run: Set all noise diodes with timestamp {} ({})'
                    .format(int(timestamp),
```
ajoubertza: In `on` and `off`, the dry-run cases are logging with `ceil` (but probably shouldn't use `int` or `ceil`)

astrokat/noisediode.py
```diff
@@ -245,11 +268,13 @@ def pattern(kat,  # kat subarray object
...
+                                                 on_fraction,
+                                                 cycle_length)
             if not kat.dry_run:
-                _katcp_reply_to_log_({ant: the_reply})
+                timestamp = _katcp_reply_to_log_({ant: the_reply})
             else:
-                msg = ('Set noise diode for antenna {} with timestamp {}'
+                msg = ('Dry-run: Set noise diode for antenna {} with timestamp {}'
                        .format(ant,
                                timestamp))
```
ajoubertza: Other dry-run logs include `ctime` as well.

PR #61, opened 2019-07-16
Display all tags

rubyvanrooyen:
List targets for all possible observation tags, ignore tags if list is empty

If multiple tags as space separated list in YAML, display all tags per target in observation log output

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -379,7 +379,10 @@ def run_observation(opts, kat):
         target_tags = []
         for cal_type in cal_tags.keys():
             target_tags.append('~{}'.format(cal_tags[cal_type]))
-            cal_list = ', '.join([repr(cal.name) for cal in catalogue.filter(cal_tags[cal_type])])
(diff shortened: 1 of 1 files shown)


astrokat/observe_main.py
```diff
@@ -370,12 +372,19 @@ def run_observation(opts, kat):
             if opts.all_up and (len(catalogue.filter(el_limit_deg=opts.horizon)) != len(catalogue)):
                 raise NotAllTargetsUpError('Not all targets are currently visible - '
                                            'please re-run the script with --visibility for information')
+        cal_tags = {'Polarisation': 'polcal',
+                    'Flux': 'fluxcal',
```
vasaantk: Hi Ruby, can you also include the 'delaycal' tag. Thanks

rubyvanrooyen: I added the "delaycal" tag, but I again advise that astronomers think about design when you put in requests. Delaycal means nothing for the observation and will lead to confusion. Ideally the processing software should be updated to reflect "gaincal" since that is the only complex gain calibration.

PR #64, opened 2019-07-19
astrokat observations log to include version info

xinyuwu:
* astrokat observations log to include version info
* added extra slewing and tracking logging to simulated log

simulated
dryrun

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -5,7 +5,6 @@ import logging
 import numpy as np
 import os
 import time
-import katversion
 
 import astrokat
 from astrokat.utility import datetime2timestamp, timestamp2datetime
@@ -238,7 +237,7 @@ class Telescope(object):
 
     def __enter__(self):
         # Verify subarray setup correct for observation before doing any work
-        user_logger.warn('Running astrokat version - {}'.format(katversion.get_version()))
+        user_logger.warn('Running astrokat version - {}'.format(astrokat.__version__))
         if 'instrument' in self.opts.obs_plan_params.keys():
             self.subarray_setup(self.opts.obs_plan_params['instrument'])
 
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -5,6 +5,7 @@ import logging
 import numpy as np
 import os
 import time
+import katversion
 
 import astrokat
 from astrokat.utility import datetime2timestamp, timestamp2datetime
@@ -676,6 +677,8 @@ def main(args):
         user_logger.setLevel(logging.DEBUG)
     if opts.trace:
         user_logger.setLevel(logging.TRACE)
+
+    user_logger.warn('Running astrokat version - {}'.format(katversion.get_version()))
 
     # setup and observation
     with Telescope(opts) as kat:
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -237,7 +237,7 @@ class Telescope(object):
 
     def __enter__(self):
         # Verify subarray setup correct for observation before doing any work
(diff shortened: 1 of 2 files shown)


astrokat/observe_main.py
```diff
@@ -237,6 +238,7 @@ def __init__(self, opts, correlator=None):
 
     def __enter__(self):
         # Verify subarray setup correct for observation before doing any work
+        user_logger.warn('Running astrokat version - {}'.format(katversion.get_version()))
```
samtwum: I think this is better logged as `info` instead of `warn`

So like you were explaining, the log didn't appear when you put it in `__init__`?

xinyuwu: > I think this is better logged as `info` instead of `warn`. Have a look at this: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels
> 
Strictly speaking, it is not a warning. However, warnings show up as a different colour which makes it more visible.

> So like you were explaining, the log didn't appear when you put it in `__init__`?
No, it only seems to work in __enter__ for dryrun. It has something to do with how katcorelib initialises logger.

ajoubertza: I agree with @samtwum that this should be info level, not warning.  It isn't special.  In normal operation the user shouldn't care much about the version - it is just there for auditing later.

ajoubertza: In the screenshot from the CAM dry-run, the version doesn't look right.  It shows as `0.0...`.  We need to investigate this.  Probably to do with how it is installed, rather than this code, but we need to resolve it.

astrokat/simulate.py
```diff
@@ -156,15 +156,20 @@ def _fake_slew_(self, target):
...
         self.track_ = True
-        time.sleep(self._fake_slew_(target)+duration)
+        time.sleep(self._fake_slew_(target))
         now = timestamp2datetime(self.time)
         simobserver.date = ephem.Date(now)
+        user_logger.info('Slewed to {}'.format(target.name))
+        time.sleep(duration)
+        now = timestamp2datetime(self.time)
+        simobserver.date = ephem.Date(now)
+        user_logger.info('Tracked {} for {}'.format(target.name, duration))
```
ajoubertza: These two logs don't quite match what session outputs.  I thought the plan was for astrokat dry-run output to be as close as possible to what CAM will output.  See what session does [here](https://github.com/ska-sa/katcorelib/blob/bc4226fe8dcd433f8dfe0485d3cf0de7c436dbb7/katcorelib/mkat_session.py#L2383).

astrokat/observe_main.py
xinyuwu: OK, fixed logging output. CAM dry-run logging output is now correct.
Offline verify logging does not include branch number, but does include commit number, eg: 0.1.dev369+head.e851323.

astrokat/simulate.py
xinyuwu: We though session logging is too verbose. The messages left out did not add any timing information.

astrokat/observe_main.py
```diff
@@ -237,6 +237,7 @@ def __init__(self, opts, correlator=None):
 
     def __enter__(self):
         # Verify subarray setup correct for observation before doing any work
+        user_logger.info('Running astrokat version - {}'.format(astrokat.__version__))
```
tockards: We tend to stick to the old `%-style of string formatting` with logging and append the variable data as arguments.
But this also works. 

e. g
`        user_logger.info('Running astrokat version - %s', astrokat.__version__)`

ajoubertza: +1 for `%` formatting of log messages.

PR #65, opened 2019-07-30
Remove hard coded cal tags

rubyvanrooyen:
Remove hardcoded dictionary of calibrator tags
Updated with code that extract all unique cal tags and use those to display output to user, thus whatever is in the YAML is displayed, making the code independent of user defined tags

New output:
```
2019-07-30 07:03:41Z - Found 5 target(s): 0 from 0 catalogue(s), 0 from default catalogue and 37 as target string(s)
2019-07-30 07:03:41Z - BP calibrators are ['J0408-6545']
2019-07-30 07:03:41Z - DELAY calibrators are ['J0408-6545']
2019-07-30 07:03:41Z - FLUX calibrators are ['J0408-6545']
2019-07-30 07:03:41Z - GAIN calibrators are ['J0632+1022', 'J1150-0023']
2019-07-30 07:03:41Z - Observation targets are ['3C 175', '3C 274']
```

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -283,8 +283,8 @@ class Telescope(object):
             return
         approved_sb_sensor_value = approved_sb_sensor.get_value()
         if self.array.sb_id_code not in approved_sb_sensor_value:
-            user_logger.info('Skipping instrument checks - {} not in approved_schedule'
-                             .format(self.array.sb_id_code))
+            user_logger.info('Skipping instrument checks - {} not in approved_schedule'.format(
+                self.array.sb_id_code))
             return
 
         for key in instrument.keys():
@@ -334,6 +334,11 @@ def run_observation(opts, kat):
     if 'durations' in obs_plan_params:
         if 'obs_duration' in obs_plan_params['durations']:
             obs_duration = obs_plan_params['durations']['obs_duration']
+    # check for nonsensical observation duration setting
+    if obs_duration < 1e-5:
+        user_logger.error('Unexpected value: obs_duration: {}'.format(
+            obs_duration))
+        return
 
     # Each observation loop contains a number of observation cycles over LST ranges
     # For a single observation loop, only a start LST and duration is required
@@ -347,6 +352,7 @@ def run_observation(opts, kat):
         target_list = obs_targets['target'].tolist()
         # build katpoint catalogues for tidy handling of targets
         catalogue = collect_targets(kat.array, target_list)
+        obs_tags = []
         for tgt in obs_targets:
             # catalogue names are no longer unique
             name = tgt['name']
@@ -356,7 +362,9 @@ def run_observation(opts, kat):
                 if (name == cat_tgt.name and
                         tags == ' '.join(cat_tgt.tags)):
                     tgt['target'] = cat_tgt
+                    obs_tags.extend(cat_tgt.tags)
                     break
+        cal_tags = np.unique([tag for tag in obs_tags if 'cal' in tag])
 
         # observer object handle to track the observation timing in a more user friendly way
         observer = catalogue._antenna.observer
@@ -376,24 +384,20 @@ def run_observation(opts, kat):
             if opts.all_up and (len(catalogue.filter(el_limit_deg=opts.horizon)) != len(catalogue)):
                 raise NotAllTargetsUpError('Not all targets are currently visible - '
                                            'please re-run the script with --visibility for information')
-        cal_tags = {'Polarisation': 'polcal',
-                    'Flux': 'fluxcal',
-                    'Bandpass': 'bpcal',
-                    'Gain': 'gaincal',
-                    'Delay': 'delaycal',
-                    }
-        target_tags = []
-        for cal_type in cal_tags.keys():
-            target_tags.append('~{}'.format(cal_tags[cal_type]))
-            cal_array = [repr(cal.name) for cal in catalogue.filter(cal_tags[cal_type])]
+        # List sources and their associated functions from observation tags
+        not_cals_filter_list = []
+        for cal_type in cal_tags:
+            not_cals_filter_list.append('~{}'.format(cal_type))
+            cal_array = [repr(cal.name) for cal in catalogue.filter(cal_type)]
             if len(cal_array) < 1:
                 continue  # do not display empty tags
             cal_list = ', '.join(cal_array)
             user_logger.info("{} calibrators are [{}]".format(
-                             cal_type,
+                             str.upper(cal_type.replace('cal', '')),
                              cal_list))
-        user_logger.info('Imaging targets are [{}]'.format(
-                         ', '.join([repr(target.name) for target in catalogue.filter(target_tags)])))
+        user_logger.info('Observation targets are [{}]'.format(
+            ', '.join([repr(target.name) for target in catalogue.filter(
+                not_cals_filter_list)])))
 
         # TODO: the description requirement in sessions should be re-evaluated
         # since the schedule block has the description
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -334,11 +334,6 @@ def run_observation(opts, kat):
     if 'durations' in obs_plan_params:
         if 'obs_duration' in obs_plan_params['durations']:
             obs_duration = obs_plan_params['durations']['obs_duration']
-    # check for nonsensical observation duration setting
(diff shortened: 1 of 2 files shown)


astrokat/observe_main.py
```diff
@@ -376,24 +379,20 @@ def run_observation(opts, kat):
...
+        not_cals_filter_list = []
+        for cal_type in cal_tags:
+            not_cals_filter_list.append('~{}'.format(cal_type))
+            cal_array = [repr(cal.name) for cal in catalogue.filter(cal_type)]
             if len(cal_array) < 1:
                 continue  # do not display empty tags
             cal_list = ', '.join(cal_array)
             user_logger.info("{} calibrators are [{}]".format(
-                             cal_type,
+                             str.upper(cal_type.replace('cal', '')),
```
ajoubertza: Hopefully no one wants to use a "calcal" type

astrokat/observe_main.py
```diff
@@ -376,24 +379,20 @@ def run_observation(opts, kat):
...
-            cal_array = [repr(cal.name) for cal in catalogue.filter(cal_tags[cal_type])]
+        # List sources and their associated functions from observation tags
+        not_cals_filter_list = []
+        for cal_type in cal_tags:
+            not_cals_filter_list.append('~{}'.format(cal_type))
+            cal_array = [repr(cal.name) for cal in catalogue.filter(cal_type)]
             if len(cal_array) < 1:
                 continue  # do not display empty tags
             cal_list = ', '.join(cal_array)
             user_logger.info("{} calibrators are [{}]".format(
```
ajoubertza: No need to change, but making a CSV string from the `cal_array` and printing inside square brackets is exactly what you get for free just using `cal_array` in the format directly.

xinyuwu:
> Code looks OK, but unit tests need updating.

Tests have been updated.

astrokat/observe_main.py
```diff
@@ -356,7 +357,9 @@ def run_observation(opts, kat):
                 if (name == cat_tgt.name and
                         tags == ' '.join(cat_tgt.tags)):
                     tgt['target'] = cat_tgt
+                    obs_tags.extend(cat_tgt.tags)
                     break
+        cal_tags = np.unique([tag for tag in obs_tags if 'cal' in tag])
```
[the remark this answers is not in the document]
rubyvanrooyen: `set()` is better than numpy -- replaced `np.unique` with `set`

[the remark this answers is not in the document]
rubyvanrooyen: My standard assumption is that calibrators will have the "cal" substring somewhere in the tag. I try not to make any assumption beyond that, because I do not know if the position of it will change, there is no clear definition on how the astronomers want this.
History shows a format `tag=<caltype>+"cal"`
But I do not know if I should enforce that, and if it is simply convenient at the moment.....

So I assume we could make an executive decision now?
I am open to suggestions...

astrokat/observe_main.py
rubyvanrooyen: Thru, perhaps I wanted to be to verbose and nice
```
2019-08-01 06:55:49Z - BP calibrators are ['J0408-6545']
2019-08-01 06:55:49Z - FLUX calibrators are ['J0408-6545']
2019-08-01 06:55:49Z - GAIN calibrators are ['J0632+1022', 'J1150-0023']
2019-08-01 06:55:49Z - DELAY calibrators are ['J0408-6545']
2019-08-01 06:55:49Z - Observation targets are ['3C 175', '3C 274']
```

What if I make it sort, with zero assumptions, only outputting the tags and information as listed?
```
2019-08-01 06:55:49Z - BPCAL ['J0408-6545']
2019-08-01 06:55:49Z - FLUXCAL ['J0408-6545']
2019-08-01 06:55:49Z - GAINCAL ['J0632+1022', 'J1150-0023']
2019-08-01 06:55:49Z - DELAYCAL ['J0408-6545']
2019-08-01 06:55:49Z - TARGETS ['3C 175', '3C 274']
```

astrokat/observe_main.py
rubyvanrooyen: Please help, I need some advice on string formatting

If I do the CSV string in brackets, I get output:
`2019-08-01 06:55:49Z - GAINCAL ['J0632+1022', 'J1150-0023']`

If I just do
`user_logger.info("{} {}".format(str.upper(cal_type), cal_array))`
I get
`2019-08-01 06:55:49Z - GAINCAL ["'J0632+1022'", "'J1150-0023'"]`

astrokat/observe_main.py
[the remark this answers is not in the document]
ajoubertza: I like `tag=<caltype>+"cal"`.  Keeps things consistent, and is still very flexible.

Who consumes the calibration tags?  SDP pipeline?  Maybe @ludwigschwardt has a strong opinion.

astrokat/observe_main.py
ajoubertza: The verbose printout is nicer.

astrokat/observe_main.py
ajoubertza: Hmmm.  Try using `str` instead of `repr`:  
`cal_array = [str(cal.name) for cal in catalogue.filter(cal_type)]`

astrokat/observe_main.py
ajoubertza: And sticking with verbose means @xinyuwu doesn't need to fix the tests again.

astrokat/observe_main.py
rubyvanrooyen: A, thank you for pointing out the obvious.
I should simply remove the `repr`, the cal.name parameter is a string (duh)

astrokat/observe_main.py
```diff
@@ -283,8 +283,8 @@ def subarray_setup(self, instrument):
             return
         approved_sb_sensor_value = approved_sb_sensor.get_value()
         if self.array.sb_id_code not in approved_sb_sensor_value:
-            user_logger.info('Skipping instrument checks - {} not in approved_schedule'
-                             .format(self.array.sb_id_code))
+            user_logger.info('Skipping instrument checks - {} not in approved_schedule'.format(
```
ludwigschwardt: Tip: loggers like the old-school formatting more, since you can go:
```Python
user_logger.info('Skipping instrument checks - %s not in approved_schedule', self.array.sb_id_code)
```
(plus the string is only formatted if the log message is actually emitted, saving a small amount of effort)

astrokat/observe_main.py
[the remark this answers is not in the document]
ludwigschwardt: The SDP pipeline consumes them (and soon katdal too). We are getting to the point where they can be formalised. It is a good assumption that the tag will end in `cal`.

astrokat/observe_main.py
ludwigschwardt: There is also the elusive `'target'` tag that may or may not become mandatory in the near future...

rubyvanrooyen: Executive decision:
If we stick with the **assumption** that calibrators will be tagged using the current format
`tag=<caltype>+"cal"`
Then the potential accidental "cal" substring from a new `caltype` name (somewhere in the future) will be inconsequential and the format definition can be applied throughout, such as
`user_logger.info("{} calibrators are {}".format(str.upper(cal_type[:-3]), cal_array))`
Providing the verbose output as previous, but not having issues with "calcal" (if the user wants to), or any naming that may contain the "cal" substring.

rubyvanrooyen:
@ludwigschwardt  'taget' tags are added in the YAML for AstroKAT
If not specified as a calibrator tag, then it carries a 'target' tag.
That is in preparation that targets in future will be more that simply radec

ludwigschwardt:
Thanks, that's useful to know (although I'm not sure why this needs a restriction to `radec`).

rubyvanrooyen:
At the moment, the input target coordinates assume `radec`
I know `katpoint` handles `altaz` and `gal` as well
I know `altaz` works in AstroKAT, but I still just need to verify that `gal` (galactic coordinates)  work as well using the YAML
Hoping we can get this in before the next open call

astrokat/observe_main.py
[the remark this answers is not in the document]
rubyvanrooyen: Implemented the format `tag=<caltype>+"cal"`

If anything happens in the future that breaks this assumption, it can be dealt with then

PR #66, opened 2019-07-30
add check for obs_duration=0

rubyvanrooyen:
Add check to exit if obs_duration is zero

New output:
```
> astrokat-observe.py --yaml tagtest.yaml
2019-07-30 14:44:54Z - Setting up telescope for observation
2019-07-30 14:44:54Z - Running astrokat version - 0.0+unknown.201907301644
2019-07-30 14:44:54Z - Request: Switch noise-diode off at 1564497900.0
2019-07-30 14:44:54Z - Dry-run: Set all noise diodes with timestamp 1564497900.0 (Tue Jul 30 16:45:00 2019)
2019-07-30 14:44:54Z - Unexpected value: obs_duration: 0.0
2019-07-30 14:44:54Z - Returning telescope to startup state
2019-07-30 14:44:54Z - Request: Switch noise-diode off at 1564497900.0
2019-07-30 14:44:54Z - Dry-run: Set all noise diodes with timestamp 1564497900.0 (Tue Jul 30 16:45:00 2019)
```

### DIFF
--- a/astrokat/utility.py
+++ b/astrokat/utility.py
@@ -28,7 +28,7 @@ def read_yaml(filename):
         return {}
 
     # remove empty keys
-    for key in data.keys():
+    for key in list(data.keys()):
         if data[key] is None:
             del data[key]
 
--- a/astrokat/utility.py
+++ b/astrokat/utility.py
@@ -27,6 +27,11 @@ def read_yaml(filename):
         # not a yaml file, suspected csv file, returning False
         return {}
 
+    # remove empty keys
+    for key in data.keys():
+        if data[key] is None:
+            del data[key]
+
     # handle mapping of user friendly keys to CAM resource keys
     if 'instrument' in data.keys():
         instrument = data['instrument']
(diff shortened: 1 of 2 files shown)


astrokat/utility.py
```diff
@@ -27,6 +27,11 @@ def read_yaml(filename):
         # not a yaml file, suspected csv file, returning False
         return {}
 
+    # remove empty keys
+    for key in data.keys():
+        if data[key] is None:
+            del data[key]
```
ajoubertza: This works in Python 2, but not in Python 3.  In general, it is a bad idea to change the container that is being iterated over.   Changing the loop to use `for key in list(data.keys()):` should work better.

Example of failure in Python 3
```
$ ipython
Python 3.6.3 (default, Oct  4 2017, 06:09:05) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: d=dict(a=1,b=None)

In [2]: for key in d.keys():
   ...:     if d[key] is None:
   ...:         del d[key]
   ...:         
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-2-fdab914f0c57> in <module>()
----> 1 for key in d.keys():
      2     if d[key] is None:
      3         del d[key]
      4 

RuntimeError: dictionary changed size during iteration

In [3]: 
```

rubyvanrooyen: Sorry, bad implementation, corrected

PR #72, opened 2019-10-25
Improve slew time estimation for simulated observations

bngcebetsha:
Adjsuting simulation slew time to be that of an ideal telescope system

[MT-1034](https://skaafrica.atlassian.net/browse/MT-1034)


astrokat/simulate.py
```diff
@@ -287,27 +288,43 @@ def scan(
...
+
         """
-        slew_speed = 2.0  # degrees / sec
-        self.katpt_current.body.compute(self.katpt_current.antenna.observer)
-        target.body.compute(target.antenna.observer)
-        try:
-            separation_angle = ephem.separation(self.katpt_current.body, target.body)
-        # TODO: need to find a clean implementation with
-        # ephem_extra.StationaryBody
-        except TypeError:
```
ajoubertza: Is this exception no longer possible?  Do the `azel` calculations not use `body.compute` internally?

astrokat/simulate.py
```diff
@@ -287,27 +288,43 @@ def scan(
...
+        el1 = katpoint.rad2deg(el1)
+        az2 = katpoint.rad2deg(az2)
+        el2 = katpoint.rad2deg(el2)
+
+        az_dist = numpy.abs(az2 - az1)
+        el_dist = numpy.abs(el2 - el1)
+
+        az_dist = az_dist if az_dist < 180. else 360. - az_dist
+
+        slew_time = max(0.5 * az_dist, 1.0 * el_dist) + 3
```
ajoubertza: Suggest some local variables to make this easier to read.  Or move the comment about the slew speeds in azimuth and elevation closer.
```python
az_speed = 2.0  # deg/sec
el_speed = 1.0  # deg/sec
overhead = 3.0  # sec
slew_time = max(az_dist / az_speed, el_dist / el_speed) + overhead
```

astrokat/simulate.py
```diff
@@ -287,27 +288,43 @@ def scan(
...
+
         return slew_time
```
ajoubertza: Earlier today there was some discussion about including the target azimuth and elevation in the user log lines, so it was being returned here.  What happened to that idea?

astrokat/simulate.py
```diff
@@ -287,27 +288,43 @@ def scan(
...
-        How long in seconds, it took to the antennas to move
-        from previous target to the next
+        How long in seconds, it took for the antennas to move
+        from current target to the next.
+
+        Notes:
+            The slew time is calculated with consideration of az-el motion
+            of antennas instead of the current angular distance between sources,
+            assuming a slew_speed 2 deg/s. Antennas slew at 2 deg/s in az while
+            moving at 1 deg/s in el. There is about 3 seconds for overhead
```
ajoubertza: Is the `SIM_OVERHEAD` and this 3 seconds due to two different effects?  I'm wondering why we need to split the delays into two different places.

tox.ini
```diff
@@ -9,3 +9,4 @@ deps =
     nosexcover
     mock
     coverage
+    unittest2
```
ajoubertza: No code changed that uses unittest2 - why did we need to add this?

astrokat/test/test_offline_observe.py
```diff
@@ -159,4 +159,4 @@ def test_image_sim(self):
         # do no need to super acurate with this target to allow
         # for slew time discrepancies
         self.assertIn("T4R02C02 observed", result)
-        self.assertIn("T4R02C04 observed for 180.0 sec", result)
+        self.assertIn("T4R02C04 observed for 360.0 sec", result)
```
ajoubertza: Will we need to update the MKAT AQF tests?

tox.ini
amakhaba: @mmphego please see the above comment

mmphego: Tox broke when we were debugging, complaining about missing unittest2

It's being imported by https://github.com/ska-sa/astrokat/blob/579161e0f30a2556ed45d043b4f28fbd18280d14/astrokat/test/test_offline_observe.py#L5 and https://github.com/ska-sa/astrokat/blob/579161e0f30a2556ed45d043b4f28fbd18280d14/astrokat/test/test_nd_timings.py#L5

but never installed, unless one runs `setup.py tests` as defined in https://github.com/ska-sa/astrokat/blob/master/setup.py#L41

ajoubertza: OK

mmphego: But since, we only needed it when we were debugging tests, do you think it's necessary to add it in tox.ini?

Also, I do not think we are being efficient. Why am I saying this, we are adding test dependencies both under `tox.ini` as well as `setup.py`.

mmphego: @amakhaba or @bngcebetsha perhaps remove test dependencies in https://github.com/ska-sa/astrokat/blob/master/setup.py#L41 and append them under `tox.ini`.

You might also consider updating `tox.ini`
see: https://github.com/ska-sa/katlogger/blob/master/tox.ini

astrokat/simulate.py
bngcebetsha: Yes they are from different effects:

- `SIM_OVERHEAD` should be named correctly, it is the time between the log line:
2019-10-17 02:07:20Z - Initialising Track gaincal J0240-2309 for 120.0 sec
And this log line
`2019-10-17 02:07:XXZ - Slewing to ...`   # This has been removed in simulations
2019-10-17 02:07:42Z - Slewed to J0240-2309

Simulation log looks like this for now:

2019-10-17 02:07:20Z - Initialising Track gaincal J0240-2309 for 120.0 sec
2019-10-17 02:07:42Z - Slewed to J0240-2309

astrokat/simulate.py
```diff
@@ -17,7 +17,7 @@
 simobserver = ephem.Observer()
 
 _DEFAULT_SLEW_TIME = 45.0  # [sec]
-SIM_OVERHEAD = 3
+_SIM_OVERHEAD = 3           # [sec]
```
mmphego: Comments should be 2 spaces are variable.
```suggestion
_SIM_OVERHEAD = 3  # [sec]
```

.gitignore
```diff
@@ -7,3 +7,6 @@
 nosetests.xml
 .idea/
 .pyenv*/
+
```
mmphego: Perhaps copy and paste gitignore from https://www.gitignore.io/api/python

bngcebetsha: It doesn't seem to have
`venv/` and `venv2/`,..the things I wanted git to ignore

astrokat/simulate.py
bngcebetsha: They do not use `body.compute`

.gitignore
mmphego: The one concern I have with this is that someone else would want to add `venv3`, `.venv`, `pyenv` etc for example.

Perhaps, only add the generic `.venv` or `venv` on the gitignore file.

Not sure if what I am saying makes any sense.

Alternatively, use this generic gitignore: https://www.gitignore.io/api/cmake,python,dotenv,virtualenv,sublimetext,visualstudiocode

astrokat/test/test_offline_observe.py
amakhaba: @ajoubertza I doubt i had a look at the AQF tests it seems like slew time is not being tested there.

PR #80, opened 2020-03-10
Scan/track cannot end below horizon limit

xinyuwu:
[MT-1281: Scan cannot end below horizon limit](https://skaafrica.atlassian.net/browse/MT-1281)
Need to make sure the target is above horizon at the end of scan or track

Also included [MT-1325: astrokat does not handle dump_rate of 0.75](https://skaafrica.atlassian.net/browse/MT-1325)

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -214,10 +214,19 @@ def cadence_target(target_list):
     return False
 
 
-def above_horizon(katpt_target, horizon=20.0):
+def above_horizon(katpt_target, horizon=20.0, duration=0.0):
     """Check target visibility."""
     [azim, elev] = katpt_target.azel(time.time())
-    return elev >= ephem.degrees(str(horizon))
+    if not elev > ephem.degrees(str(horizon)):
+        return False
+
+    if duration:
+        x = time.time()
+        y = time.time() + duration
+        [azim, elev] = katpt_target.azel(time.time() + duration)
+        return elev > ephem.degrees(str(horizon))
+
+    return True
 
 
 class Telescope(object):
@@ -534,7 +543,10 @@ def run_observation(opts, kat):
                         "TRACE: initial observer for target\n {}".format(observer)
                     )
                     # check target visible before doing anything
-                    if not above_horizon(katpt_target, horizon=opts.horizon):
+                    # make sure the target would be visible for the entire duration
+                    target_duration = target['duration']
+                    if not above_horizon(katpt_target, horizon=opts.horizon,
+                                         duration=target_duration):
                         show_horizon_status = True
                         # warning for cadence targets only when they are due
                         if (
@@ -580,7 +592,8 @@ def run_observation(opts, kat):
                             "TRACE: target observation # {} last observed "
                             "{}".format(tgt["obs_cntr"], tgt["last_observed"])
                         )
-                        if above_horizon(catalogue[tgt["name"]], horizon=opts.horizon):
+                        if above_horizon(catalogue[tgt["name"]], horizon=opts.horizon,
+                                         duration=tgt["duration"]):
                             if observe(session, tgt, **obs_plan_params):
                                 targets_visible += True
                                 tgt["obs_cntr"] += 1
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -29,6 +29,10 @@ except ImportError:
     from astrokat import collect_targets, user_logger, start_session, verify_and_connect
 
 
+# Maximum difference allowed between requested and actual dump rate (Hz)
+DUMP_RATE_TOLERANCE = 0.002
+
+
 # TODO: target description defined in function needs to be in configuration
 def read_targets(target_items):
     """Read targets info.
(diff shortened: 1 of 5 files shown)


astrokat/observe_main.py
```diff
@@ -534,7 +545,10 @@ def run_observation(opts, kat):
                         "TRACE: initial observer for target\n {}".format(observer)
                     )
                     # check target visible before doing anything
-                    if not above_horizon(katpt_target, horizon=opts.horizon):
+                    # make sure the target would be visible for the entire duration
+                    target_duration = target['duration']
```
ajoubertza: This duration excludes the time taken to observe the cadence targets, which happens before `target` is observed.  If it takes "long" for the cadence targets to be observed, then `target` may set during its actual observation, even though the check now says it will be up.

Is the time to observe cadence targets significant, i.e., "long"?  I guess that it would vary widely, depending what the astronomer had in mind.

We could add another check for `above_horizon` after the cadence targets loop to be more accurate.  This would only help for astrokat simulation (see next paragraph).

In CAM executions using session the `track`, `scan` and `raster_scan` methods include a check for targets above the horizon, while the astrokat simulation implementation of those method does not.  This will still give slightly different results.

astrokat/test/test_offline_observe.py
```diff
@@ -161,3 +161,32 @@ def test_image_sim(self):
         # for slew time discrepancies
         self.assertIn("T4R02C02 observed", result)
         self.assertIn("T4R02C04 observed for 360.0 sec", result)
+
+    def test_below_horizon(self):
+        """Test image sim."""
```
ajoubertza: Please update copy-pasted docstring.

astrokat/test/test_offline_observe.py
```diff
@@ -161,3 +161,32 @@ def test_image_sim(self):
...
+        self.assertIn(
+            expected_results, result, "MAXIJ1810-22 skipped"
+        )
+
+        # J1833-2103 with cadence started off above horizon, but at end of the duration,
+        # it would be below horizon
+        expected_results = 'Target J1833-2103 below 20.0 deg horizon, continuing'
+        self.assertIn(
+            expected_results, result, "J1833-2103 skipped"
+        )
```
ajoubertza: Thanks for including a test.

astrokat/utility.py
```diff
@@ -41,7 +41,10 @@ def read_yaml(filename):
         if instrument is not None:
             if "integration_time" in instrument.keys():
                 integration_time = float(instrument["integration_time"])
-                instrument["dump_rate"] = 1.0 / integration_time
+                if integration_time == 1.33:
+                    instrument["dump_rate"] = 0.75
```
ajoubertza: These magic values are not ideal - likely to require future maintenance, and it can easily be avoided.  We can stick with the old code here, and just fix the comparison in `Telescope.subarray_setup`, which is the only place it seems to be used.

astrokat/observe_main.py
```diff
@@ -312,7 +319,11 @@ def subarray_setup(self, instrument):
             conf_param = instrument[key]
             if key == "integration_time":
                 key = "dump_rate"
-                instrument[key] = 1.0 / float(conf_param)
+                integration_time = float(conf_param)
+                if integration_time == 1.33:
+                    instrument[key] = 0.75
+                else:
+                    instrument[key] = 1.0 / integration_time
```
ajoubertza: The `elif conf_param != sub_sensor:` line (348), was giving the problem when comparing the dump rate due to variations in precision.  We already have some special cases based on the key, so why not just add another another for "dump_rate"?  In general, comparing floating point numbers with `==` is not a reliable approach.  Especially if those numbers are calculated in different systems.  It is quite likely that there is some tiny difference. 

```python
# Maximum difference allowed between requested and actual dump rate (Hz)
DUMP_RATE_TOLERANCE = 0.002
...

            elif key == "dump_rate":
                delta = abs(conf_param - sub_sensor)
                if delta > DUMP_RATE_TOLERANCE: 
                    raise RuntimeError(
                       "Subarray configuration {} error, {} required, "
                        "{} found, delta > tolerance ({} > {})".format(
                            sensor_name, conf_param, sub_sensor, delta, DUMP_RATE_TOLERANCE)
                    )
            elif conf_param != sub_sensor:
                raise RuntimeError(
                    "Subarray configuration {} error, {} required, "
                    "{} found".format(sensor_name, conf_param, sub_sensor)
                )
```

I'd prefer a tolerance of 0.001 Hz, but then the OPT needs to use an integration time of 1.333 seconds instead of 1.33.  Is that possible?

astrokat/observe_main.py
```diff
@@ -312,7 +319,11 @@ def subarray_setup(self, instrument):
             conf_param = instrument[key]
             if key == "integration_time":
```
ajoubertza: Can't we remove this whole `if key == "integration_time"` block?  The yaml file loader always replaces the "integration_time" key with "dump_rate", so how would we end up with "integration_time" here?

astrokat/observe_main.py
xinyuwu: It is correct that it does not include cadence observation time. But cadence is no longer supported by OPT, and the fix to include cadence is complicated.

astrokat/test/test_offline_observe.py
xinyuwu: yes, done.

astrokat/utility.py
xinyuwu: done

astrokat/observe_main.py
xinyuwu: done

astrokat/observe_main.py
xinyuwu: done

PR #83, opened 2020-04-01
use local copy of ephem target object

rubyvanrooyen:
So the issue is actually ephem and python 2.7

The python 2 versions used pass by references, so changing the local copy in a function will also update the original.
(this is fixed in python 3)

I fixed this by making explicit local copies -- this should work in both 2.7 and 3


rubyvanrooyen:
Test on devcomm
http://10.8.67.30:8081/tailtask/20200401-0001/dryrun

astrokat/observe_main.py
```diff
@@ -220,14 +220,34 @@ def cadence_target(target_list):
...
 def above_horizon(katpt_target, horizon=20.0, duration=0.0):
     """Check target visibility."""
-    [azim, elev] = katpt_target.azel(time.time())
+    # use local copies so you do not overwrite target time attribute
+    if 'azel' in katpt_target.tags:
+        return True
+
+    # must be celestial target (ra, dec)
+    local_target = katpt_target.body.copy()
+    local_observer = katpt_target.antenna.observer.copy()
```
ajoubertza, line 229: In katpoint's [Target.azel](https://github.com/ska-sa/katpoint/blob/e6a9efdff0c2b4d87e33e0a4a8a1f81117ea7117/katpoint/target.py#L310) method, we can pass an antenna object.  If that object is a copy, then it will modify that copy instead of the original.

So something like this:
```
local_antenna = copy.deepcopy(katpt_target.antenna)
azim, elev = katpt_target.azel(time.time(), antenna=local_antenna)
...
azim, elev = katpt_target.azel(time.time() + duration, antenna=local_antenna)
```

astrokat/observe_main.py
```diff
@@ -220,14 +220,34 @@ def cadence_target(target_list):
...
 
+    # check that target will be visible at end of track
     if duration:
-        [azim, elev] = katpt_target.azel(time.time() + duration)
-        # dummy call to reset timestamp for logger since azel() changes timestampe
-        [x, y] = katpt_target.azel(time.time())
+        end_ = timestamp2datetime(time.time() + duration)
+        local_observer.date = ephem.date(end_)
+        local_target.compute(local_observer)
+        [azim, elev] = [local_target.az, local_target.alt]
```
ajoubertza, line 247: This is almost identical to lines 232 to 235.  Suggest the common code is moved to a function, so we don't have to repeat ourselves.  
(Or use the katpoint.azel method)

astrokat/observe_main.py
```diff
@@ -220,14 +220,34 @@ def cadence_target(target_list):
 
 def above_horizon(katpt_target, horizon=20.0, duration=0.0):
     """Check target visibility."""
-    [azim, elev] = katpt_target.azel(time.time())
+    # use local copies so you do not overwrite target time attribute
+    if 'azel' in katpt_target.tags:
+        return True
```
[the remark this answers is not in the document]
rubyvanrooyen: Sorry, incomplete test -- updated to include desired functionality
```
if katpt_target.body.alt < horizon:
    return False
else:
    return True
```

Since altitude is given all that is needed is a direct check
No date computation is needed, function exits quickly

astrokat/observe_main.py
rubyvanrooyen: General clean up of the `above_horizon` function implementation.
Target and observer are now passed explicitly with the observer being a copy of the antenna object.
This seems to work and keeps the `katpt_target` contained in the `run_observation` and explicitly requires ephem objects in the horizon function.

astrokat/observe_main.py
rubyvanrooyen: Cleaned up duplicate code, preference to using a utility function that explicitly requires ephem targets

astrokat/observe_main.py
```diff
@@ -218,17 +225,47 @@ def cadence_target(target_list):
...
+    if type(target) is not ephem.FixedBody:
+        # anticipate katpoint special target for AzEl targets
+        if 'alt' not in vars(target):
+            raise RuntimeError('Unknown target type, exiting...')
+        # 'StationaryBody' objects do not have RaDec coordinates
+        # check pointing altitude is above minimum elevation limit
+        if target.alt < horizon:
+            return False
+        else:
+            return True
```
mmphego, line 245: Instead of writing it like that, you could rewrite it this way.

```suggestion
        return bool(target.alt < horizon)
```

rubyvanrooyen:
@mmphego please can you assist with the py27 test failure.
I cannot re-create the error on my system.
The error reported
```
[2020-05-04T12:51:06.385Z] py27 run-test: commands[1] | coverage xml -o coverage_py27.xml
[2020-05-04T12:51:06.385Z] No data to report.
```

When I run `LC_ALL=C test_flags= tox -e py27` on my mac
```
  py27: commands succeeded
  congratulations :)
```
I attach the full output report 
[tox_out.txt](https://github.com/ska-sa/astrokat/files/4575148/tox_out.txt)

mmphego:
> @mmphego please can you assist with the py27 test failure.
> I cannot re-create the error on my system.
> The error reported
> 
> ```
> [2020-05-04T12:51:06.385Z] py27 run-test: commands[1] | coverage xml -o coverage_py27.xml
> [2020-05-04T12:51:06.385Z] No data to report.
> ```
> 
> When I run `LC_ALL=C test_flags= tox -e py27` on my mac
> 
> ```
>   py27: commands succeeded
>   congratulations :)
> ```
> 
> I attach the full output report
> [tox_out.txt](https://github.com/ska-sa/astrokat/files/4575148/tox_out.txt)

Mhhh, this is a bit odd - it works 100% on my pc and ran another build which was also successful. Might be due to the version of `coverage` we are using - I will investigate.

rubyvanrooyen:
@mmphego fixed the test error for py27, merging with master

PR #84, opened 2020-04-26
catalogue debuggings have incorrect timing.

xinyuwu:
[MT-1257: Timestamp in astrokat during simulate](https://skaafrica.atlassian.net/browse/MT-1257)

Fixing last of the simulation logging timing issue.  Cataglue does not need a date.

### DIFF
--- a/astrokat/observatory.py
+++ b/astrokat/observatory.py
@@ -263,7 +263,6 @@ def collect_targets(kat, args):
     from_names = from_strings = from_catalogues = num_catalogues = 0
     catalogue = katpoint.Catalogue()
     catalogue.antenna = katpoint.Antenna(_ref_location)
-    catalogue.antenna.observer.date = lst2utc(kat._lst, _ref_location)
 
     setobserver(catalogue.antenna.observer)
 


astrokat/observatory.py
```diff
@@ -263,7 +263,6 @@ def collect_targets(kat, args):
     from_names = from_strings = from_catalogues = num_catalogues = 0
     catalogue = katpoint.Catalogue()
     catalogue.antenna = katpoint.Antenna(_ref_location)
-    catalogue.antenna.observer.date = lst2utc(kat._lst, _ref_location)
```
ajoubertza, line 266: In simulation mode is `kat.array.dry_run` true?

In `observe_main.run_observation`, the catalogue is used to filter the targets based on elevation as an early check, if we are not doing a dry run:  

https://github.com/ska-sa/astrokat/blob/82727f493f7e2ac6651c06d658f01982ce148ff7/astrokat/observe_main.py#L415-L428

That must require a time, and a time is only set a bit later:

https://github.com/ska-sa/astrokat/blob/82727f493f7e2ac6651c06d658f01982ce148ff7/astrokat/observe_main.py#L466

If dry_run is true for simulations, then that's fine.  For real observations, we are using `collect_targets` from `katcorelib.observe`, so this change in astrokat will have no effect.

xinyuwu: Good pickup. I can confirm that for simulation kat.array.dry_run is set to true.

PR #91, opened 2020-05-25
More planning scripts

rubyvanrooyen:
Last two helper scripts to assist users offline. Both use astropy with the respective functionalities:
* `astrokat-coords.py` is a simple command line tool to convert Galactic coordinates (gal_lat, gal_lon) to equatorial coordinates (ra, dec)
* `astrokat-uvcoverage.py` give some indication of UV coverage dependent on array antennas selected


scripts/astrokat-coords.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if args.verbose:
+        print('Catalogue file entries')
+        print(cat_entry)
+
+
+if __name__ == '__main__':
+    # This code will take care of negative values for declination
+    for i, arg in enumerate(sys.argv):
+        if (arg[0] == '-') and arg[1].isdigit():
```
ajoubertza, line 190: May also raise an exception if len(arg) < 2.

config/mkat_antennas.yml
```diff
@@ -0,0 +1,69 @@
+reference:
+  latitude: '-30:42:47.4'
+  longitude: '21:26:38.0'
+  altitude: 1060.0
```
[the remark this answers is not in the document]
rubyvanrooyen: I used values from original ground measurements.
I updated the values to those in the config file

config/mkat_antennas.yml
```diff
@@ -0,0 +1,69 @@
...
+antennas:
+  - name=m000, diameter=13.5, north=-8.21, east=26.797, up=0.117
```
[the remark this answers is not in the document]
rubyvanrooyen: I updated the values to be the most recent as per the delay config files.
I also added a helper script for doing the update in future, since the delay values are continually updating.

scripts/astrokat-coords.py
rubyvanrooyen: Very good suggestion, much appreciated
I updated all the scripts that use this little hack to add this extra test

rubyvanrooyen:
@ajoubertza , @xinyuwu 

I updated the UV coverage script:
* cleaned up all the trailing units from astropy that caused unexpected errors
* separated the main simulation functionality to take specific parameters -- this should make it easier for you to call the main function from another script
* it now uses a target coordinate and observation duration to calculate the hour angle range

rubyvanrooyen:
@ajoubertza can you please indicate if I can merge this branch with master

config/update_mkat_antennas.py
```diff
@@ -0,0 +1,103 @@
+"""Update MeerKAT antenna position file"""
```
ajoubertza, line 1: This is handy for future updates 👍

scripts/astrokat-uvcoverage.py
```diff
@@ -1,13 +1,16 @@
...
 from __future__ import print_function
 from astropy import units as u
 from astropy.coordinates import Longitude, Latitude, EarthLocation
+from astropy.coordinates import ICRS, SkyCoord
+from datetime import datetime, timedelta
 from numpy import recarray
 
 import argparse
 import astropy.constants as phys
+import ephem
```
ajoubertza, line 13: OK, but I thought this script was trying to do everything with astropy?

scripts/astrokat-uvcoverage.py
```diff
@@ -31,13 +34,15 @@ def __init__(self,
         self.wavelength = phys.c / self.frequency
 
         self.ref_position = None
-        arr_desc = {'names': ('name', 'north', 'east', 'up', 'location'),
+        self.observer = None
+        arr_desc = {'names': ('name', 'east', 'north', 'up', 'location'),
                     'formats': ('S5', float, float, float, EarthLocation)}
```
ajoubertza, line 39: Does S5 mean up to 5 chars?  I'm wondering about SKA / MKAT+ dishes, which will have names like `s0013`, in which case 5 is enough.

scripts/astrokat-uvcoverage.py
rubyvanrooyen: astropy does not have an "observer" for that you have to use astroplan, which in itself is a very good tool, but it is >py3.5, so I had to use ephem for the interim until 2.7 is fully retired
Since ephem is only used for the observer, but the targets are astropy, I will sidestep the rounding issues of ephem in pointing

scripts/astrokat-uvcoverage.py
rubyvanrooyen: Yes, s5 means max of 5 characters are read, so if you give more than 5 it truncates.
I used the meerkat array naming convention -- 5 should be sufficient.

PR #92, opened 2020-06-09
Slew instead of 0-second track for first target to help FBFUSE

rubyvanrooyen:
`FBFUSE` / MeerTIME observations had some issues with the astrokat observation script when `FBFUSE` ran commensally in the array with a MeerTIME pulsar observation.

Errors and ideas were discussed during a site integration session and it was noticed that astrokat on the "slew to first target" -- lacks the `fbf.check_provisioning_beams_complete` as implemented in `sessions.capture_start`. This allows the FBF setup of the beams

For astrokat the optimal implementation of this missing requirement was to add an explicit *slew* function to replace the current 0 sec duration track that mimicked the slew.
During this slew to first target (which is expected to be the longest slew in general) the FBF beam check is initiated once the antennas are instructed to POINT to target, this should utilise the slew dead time for FBF config

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -151,10 +151,15 @@ def initial_slew(session, target_info):
             raise session.FBFUSEError("Provision beams not ready")
 
     # Wait until a quorum is in position (with timeout)
-    if not session.kat.dry_run:
-        session.wait(session.ants, 'lock', True, timeout=300, quorum=session.quorum)
+    success = session.wait(session.ants,
+                           'lock',
+                           True,
+                           timeout=300,
+                           quorum=session.quorum)
+    if success:
+        user_logger.info('target reached')
+    # session.wait will issue a waring if quorum not reached
 
-    user_logger.info('target reached')
 
 
 def observe(session, target_info, **kwargs):
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -122,18 +122,38 @@ def read_targets(target_items):
     return target_list
 
 
-def slew(session, target):
+def slew(session, target_info):
+    """Simple way to get telescope to slew to target
+
+    Parameters
+    ----------
+    session: `CaptureSession` object
+    target_info: dictionary with target observation info
+
+    """
+    katpt_tgt = target_info["target"]
+
+    user_logger.info("Slewing to first target")
+    session.set_target(katpt_tgt)
     session.activity('slew')
-    user_logger.info('slewing to target')
     # Start moving each antenna to the target
     if session.kat.dry_run:
-        session._slew_to(target)
+        # Apply average slew time
+        session._slew_to(katpt_tgt)
     else:
-        ants = session.ants
-        # ants = session.kat.ants
-        ants.req.mode('POINT')
-        # Wait until a quorum is in position (with timeout)
-        session.wait(ants, 'lock', True, timeout=300, quorum=session.quorum)
+        # Start moving each antenna to the target
+        session.ants.req.mode('POINT')
+
+    # Initiate FBFUSE proxy settings from mkat-sessions capture_start
+    if session.fbf:
+        user_logger.info('Initiating FBF beam complete')
+        if not session.fbf.check_provisioning_beams_complete():
+            raise session.FBFUSEError("Provision beams not ready")
+
+    # Wait until a quorum is in position (with timeout)
+    if not session.kat.dry_run:
+        session.wait(session.ants, 'lock', True, timeout=300, quorum=session.quorum)
+
     user_logger.info('target reached')
 
 
@@ -142,8 +162,8 @@ def observe(session, target_info, **kwargs):
 
     Parameters
     ----------
-    session: `CaptureSession`
-    target_info:
+    session: `CaptureSession` object
+    target_info: dictionary with target observation info
 
     """
     target_visible = False
@@ -152,16 +172,6 @@ def observe(session, target_info, **kwargs):
     target = target_info["target"]
     duration = target_info["duration"]
     obs_type = target_info["obs_type"]
-
-    # simple way to get telescope to slew to target
-    if "slewonly" in kwargs:
-        # RVR test
-        # only for integration tests, remove after test
-        print('Using slew only function')
-        return slew(session, target)
-        # RVR test
-#         print('0 sec track')
-#         return session.track(target, duration=0.0, announce=False)
 
     # set noise diode behaviour
     nd_setup = None
@@ -445,21 +455,6 @@ def run_observation(opts, kat):
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
 
-    # if FBFUSE proxy is available, this observer allows it to block
-    # at startup of observation to allow setup of beams
-    fbf_block_allowed = False
-    if "instrument" in obs_plan_params:
-        instrument_dict = obs_plan_params['instrument']
-        if "partner_resources" in instrument_dict:
-            partners = instrument_dict['partner_resources']
-            if 'fbfuse' in partners:
-                fbf_block_allowed = True
-
-    # RVR test
-    # only for integration tests, remove after test
-    print('FBF block is allowed = {}'.format(fbf_block_allowed))
-    # RVR test
-
     # set up duration periods for observation control
     obs_duration = -1
     if "durations" in obs_plan_params:
@@ -642,26 +637,10 @@ def run_observation(opts, kat):
                 "DEBUG: Initialise capture start with timestamp "
                 "{} ({})".format(int(time.time()), timestamp2datetime(time.time()))
             )
-            # Adding the FBF proxy BLOCK command here
-            # Thus will allow an up to 5 minute wait to all observations
-            # that is run on an array with the FBFUSE proxy included in the
-            # array
-
-            # RVR test
-            # only for integration tests, remove after test
-            print('FBF proxy in array = {}'.format(session.fbf))
-            # RVR test
-
-            # if session.fbf and fbf_block_allowed:
-            if session.fbf:
-                user_logger.warn('Initiating FBF beam complete block')
-                if not session.fbf.check_provisioning_beams_complete():
-                    raise session.FBFUSEError("Provision beams not ready")
-                user_logger.info('Continue observation')
 
             # Go to first target before starting capture
-            user_logger.info("Slewing to first target")
-            observe(session, obs_targets[0], slewonly=True)
+            slew(session, obs_targets[0])
+
             # Only start capturing once we are on target
             session.capture_start()
             user_logger.trace(
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -146,7 +146,7 @@ def initial_slew(session, target_info):
 
     # Initiate FBFUSE proxy settings from mkat-sessions capture_start
     if session.fbf:
-        user_logger.info('Initiating FBF beam complete')
+        user_logger.info('Waiting for FBF beam complete')
         if not session.fbf.check_provisioning_beams_complete():
             raise session.FBFUSEError("Provision beams not ready")
 
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -424,6 +424,21 @@ def run_observation(opts, kat):
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
 
+    # if FBFUSE proxy is available, this observer allows it to block
+    # at startup of observation to allow setup of beams
+    fbf_block_allowed = False
+    if "instrument" in obs_plan_params:
+        instrument_dict = obs_plan_params['instrument']
+        if "partner_resources" in instrument_dict:
+            partners = instrument_dict['partner_resources']
+            if 'fbfuse' in partners:
+                fbf_block_allowed = True
+
+    # RVR test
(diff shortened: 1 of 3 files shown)


rubyvanrooyen:
Maciej and I verified functionality on site using a little TRAPUM integration time
http://10.97.1.13:8081/tailtask/20200611-0015/progress

Observations ran successfully

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+        user_logger.info('Initiating FBF beam complete')
+        if not session.fbf.check_provisioning_beams_complete():
+            raise session.FBFUSEError("Provision beams not ready")
+
+    # Wait until a quorum is in position (with timeout)
+    if not session.kat.dry_run:
+        session.wait(session.ants, 'lock', True, timeout=300, quorum=session.quorum)
+
+    user_logger.info('target reached')
+
```
ajoubertza, line 158: `session.wait` returns a boolean:  True if a quorum of antennas reached the target in time.  The user log here should be dependent on that return value.

And do we want the observation to continue if we couldn't get to the first target?

astrokat/simulate.py
```diff
@@ -160,6 +160,13 @@ def __init__(self, kat, **kwargs):
         self.time = self.start_time
         self.katpt_current = None
         self.capture_initialised = False
+        # add USE proxy indicators
+        if "instrument" in self.obs_params:
+            instrument_dict = self.obs_params['instrument']
+            if "pool_resources" in instrument_dict:
+                pool_resources = instrument_dict['pool_resources']
+                if 'fbfuse' not in pool_resources:
+                    self.fbf = False
```
ajoubertza, line 169: Do we ever set `self.fbf` to something else?  I am thinking about the case where we are simulating with fbf.  Does `if session.fbf` evaluate to true somehow?  If it is true, what code is run when we call
`if not session.fbf.check_provisioning_beams_complete()`?

Similarly, if instrument or pool resources isn't specified, will `if session.fbf` evaluate to true or false?  I would hope false, but I'm not sure how that would happen.

lanceWilliams: Technically self.fbf should only ever be the returned instance or None. There should not be anything that sets it to false. The fbfuse data proxy wrapper looks for a fbfuse attribute on the kat object. I don't know if the instrument_dict in obs_params emulates that perfectly/reliably.

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+        # Apply average slew time
+        session._slew_to(katpt_tgt)
+    else:
+        # Start moving each antenna to the target
+        session.ants.req.mode('POINT')
+
+    # Initiate FBFUSE proxy settings from mkat-sessions capture_start
+    if session.fbf:
+        user_logger.info('Initiating FBF beam complete')
+        if not session.fbf.check_provisioning_beams_complete():
```
lanceWilliams, line 150: @rubyvanrooyen  This check for beam provisioning is potentially a relatively long running blocking function. Does it not perhaps interfere with the timing that MeerTime wants after the target is set, or do they not care about this when slewing to the first target? I seem to remember their commands to trigger their backend being baked into their scripts somehow.

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+
+    Parameters
+    ----------
+    session: `CaptureSession` object
+    target_info: dictionary with target observation info
+
+    """
+    katpt_tgt = target_info["target"]
+
+    user_logger.info("Slewing to first target")
```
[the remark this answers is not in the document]
rubyvanrooyen: I agree, the naming of the function makes its purpose unclear, since the functionality only focus on the anticipated long slew to the first target.
Assumption warning:
In general moving from the last target of the previous observation to the first target of this observation will be the longest slew.

During this time AstroKAT does not capture data, it only start once on target, but then I do not want to run into the FBF wait of `capture_start`.

So I think I should fix both issues.
* I renamed the function to `initial_slew` to be explicit about the functionality.
* I also updated the info message since I like to target name feedback for the observer (thank you, good suggestion)

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+    session.set_target(katpt_tgt)
+    session.activity('slew')
+    # Start moving each antenna to the target
+    if session.kat.dry_run:
+        # Apply average slew time
+        session._slew_to(katpt_tgt)
+    else:
+        # Start moving each antenna to the target
```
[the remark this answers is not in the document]
rubyvanrooyen: Sorry, missed the comment in cleanup after adding the `dry-run` option
Removed the first comment, only have the comment in the `else` where it is appropriate

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+    ----------
+    session: `CaptureSession` object
+    target_info: dictionary with target observation info
+
+    """
+    katpt_tgt = target_info["target"]
+
+    user_logger.info("Slewing to first target")
+    session.set_target(katpt_tgt)
+    session.activity('slew')
```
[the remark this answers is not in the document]
rubyvanrooyen: Yes, this is what bad naming does for one :(
Sorry @ajoubertza 

I would have loved to have a separate `slew` option in `mkat_sessions`, but I needed the slew functionality for the initial slew in AstroKAT, so I only C&P selected code.
(Yes, I am aware how bad this style of coding is)

The `slew` functionality in AstroKAT does not mimic the general slew in `track`.
When starting a new observation, I anticipate that the longest slew will be when moving from the last target of the previous observation to the first target of this observation.
Consequently, AstroKAT does not do any data capture during this time, it is classified as dead time.

I therefore do not explicitly check if you are on target, although it is not unexpected to be on target when you start, I believe these types of cases to be in the minority.
This initial slew will be followed by a `capture_start` once you enter the observation target list, which will initiate the `track` and set the labels even for the first target.

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+    # Start moving each antenna to the target
+    if session.kat.dry_run:
+        # Apply average slew time
+        session._slew_to(katpt_tgt)
```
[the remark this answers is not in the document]
rubyvanrooyen: Forgive me, I was already uneasy about the C&P of the slew commands :(

astrokat/observe_main.py
```diff
@@ -122,13 +122,48 @@ def read_targets(target_items):
...
+    else:
+        # Start moving each antenna to the target
+        session.ants.req.mode('POINT')
+
+    # Initiate FBFUSE proxy settings from mkat-sessions capture_start
+    if session.fbf:
+        user_logger.info('Initiating FBF beam complete')
```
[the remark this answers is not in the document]
rubyvanrooyen: Good suggestion, thank you

astrokat/observe_main.py
rubyvanrooyen: Indeed an oversight, I corrected this and made the log message dependent on a return `True` value.
Since `session.wait` will issue a warning, I am not going to take any action toward interrupting the observation.
The `on_target` test in the `tract` command that will follow should either allow the quorum to be reached, or it will react appropriately.

astrokat/simulate.py
rubyvanrooyen: What I can see from our test runs.
If FBFUSE proxy is in the array, an instance is returned, else None

For the offline simulations I am not going to build fake proxies for USEs, according to me this type of interaction with MKAT observations should not exist.
That said, there seem to be a special case here and I needed a way to check if the FBF proxy was included.

0) When FBF is desired to be in the array, it has to added to the `pool_resources` to ensure the observation does not run without it. So I used this as indicator to mimic having an FBFUSE proxy
1) Next I needed a way to add this into the `simulation` code for when you do not have the `katcorelib` functionality.
Here I simply set it to `False` in the offline simulation since it reacts the same, but is the least amount of fiddle with the code.

This should not affect anything, because `simulate.py` code get's ignored on our systems when `mkat_session` is available.

astrokat/observe_main.py
rubyvanrooyen: This should never had to come up for AstroKAT and should have been fully handled between `mkat_sessions` and the proxies, but there seems to be missing specifications along with a priority need to have this functioning for observations.

I honestly do not agree with the `mkat_sessions` implementation that simply checks if the proxy is present since FBF can run in _commensal mode_ also :(
The presence of the proxy will trigger a wait in `capture_start`, while both `capture_start` and `on_target` seem to rely on this proxy to function if it is included in the array and this worries me.

Discussions with Ewan and Maciej highlighted that the FBFUSE proxy usage has a range or needs.
Since I am not part of the FBFUSE group I do not know the operational needs, but I do understand that they differ from the other USEs in that they under certain situations have the rights to "pause" MKAT observations in order to set up the beams (our current situation).
The implementation as it stands in `mkat_sessions` will impact **ALL** observations that include the FBFUSE proxy and use `capture_start`, even if FBFUSE is running as commensal.

This AstroKAT update simply tries to accommodate the new operational need for the FBF implementation, as well as try to mitigate the impact on observations by "hiding" the potential wait time in the dead time during the first slew.
Else, it will be an impact on the first target observation, which is definitely undesirable.
And the inclusion of the proxy as commensal may be affected, which will mean array build considerations for other observations that do not have this agreement with FBF.

For the moment, a short term solution must be found in order for observations to continue.
During our integration tests all the beam setups was < 90 seconds, which I believe to be acceptable enough to include in the long slew.
If you happen to be on target you will loose a little time, but generally the slew to the first target is your longest dead time during an observation and you do not care about that data, which is why AstroKAT only starts capture when on first target.

The longer term solution is up to the astronomers.
Perhaps @mserylak can advise us here.

mserylak: Hi @rubyvanrooyen and all.
Thank you for the explanation. It is exactly as you mentioned. `mkat_sessions` does not take into account FBFUSE being commensal, thus having no right to "block" the observation. This is at present the only way of circumventing this "feature" in `mkat_sessions` that was discussed between Ruby, Ewan and myself. As far as I am aware, FBFUSE does not require precision timing data for their observations.

As for FBFUSE being commensal with MeerTIME, AstroKAT is not used for these type of observations, but scripts under `katusescripts/ptuse`. And since that type of observations were performed for a while now, I do not think we need to worry. Otherwise it would be signalled to us by MeerTIME people already.

lanceWilliams: @mserylak with regards to the last paragraph, is the intention to no longer use the PTUSE scripts and use astrokat instead thus initiating this PR? I'm not sure how else you would encounter this issue unless there are other observations that need to do things involving target points before the capture_start is requested?

mserylak: @lanceWilliams I guess that would be desired. However, I am not in position to tell if/when this is going to happen. If that is the direction agreed/desired by CAM and PTUSE then I imagine some serious work will need to be done both on AstroKAT and sessions. As you said, `katusescripts/ptuse` has certain commands baked in and I do not know much about that at the momen.

rubyvanrooyen:
@mmphego I see this branch of mine has similar unit test issues than you mentioned in the Jira.
The master branch tests succeeds, but this branch does not. I cannot see a difference since I cannot see a difference in the tests with master branch.

Can you please help me to figure out what I should do to correct the unit tests :(

mserylak:
Just a gentle poke here to the CAM people :) Is there anything I could do to speed up the merge?

ajoubertza:
Hi @mserylak.

Do you want to fix the noise diode tests that are now failing due to changes in this PR?

ajoubertza:
The noise diode tests are brittle, checking for exact absolute time values in the tasklog output.  It looks like some of those values are now different due to a slight variation in timing cause by the initial slew.

ajoubertza:
This update should get deployed on site next week with the v27 CAM release.

astrokat/observe_main.py
```diff
@@ -122,13 +122,52 @@ def read_targets(target_items):
...
+        session.ants.req.mode('POINT')
+
+    # Initiate FBFUSE proxy settings from mkat-sessions capture_start
+    if session.fbf:
+        user_logger.info('Waiting for FBF beam complete')
+        if not session.fbf.check_provisioning_beams_complete():
+            raise session.FBFUSEError("Provision beams not ready")
+
+    # Wait until a quorum is in position (with timeout)
+    success = session.wait(session.ants,
```
ajoubertza, line 154: Turns out that this fails in a CAM dry-run:
```
Unhandled exception <type 'exceptions.AttributeError'> : 'TimeSession' object has no attribute 'wait'
Traceback (most recent call last):
  File "/usr/local/bin/astrokat-observe.py", line 8, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/local/lib/python2.7/dist-packages/astrokat/observe_main.py", line 952, in main
    run_observation(opts, kat)
  File "/usr/local/lib/python2.7/dist-packages/astrokat/observe_main.py", line 646, in run_observation
    initial_slew(session, obs_targets[0])
  File "/usr/local/lib/python2.7/dist-packages/astrokat/observe_main.py", line 154, in initial_slew
    success = session.wait(session.ants,
Task 20200702-0001 entered state 'FAILED' with reason: Non-zero script process exit status: 1.
```
I'll make a fix.

PR #96, opened 2021-04-19
OPS-1899: Increase noise diode lead time and handle errors better

ludwigschwardt:
Turning off 63 L-band noise diodes on 2021-04-15 took about 3.3 seconds, much longer than the usual 1.0 - 1.5 seconds. This caused the final few KATCP requests to fail and the script to crash.

The culprit is the increased number of S-band digitisers being serviced by the digitiser master controller, so this situation may persist in the near future. For now, increase the lead time from 3.0 to 5.0 seconds.

Do some refactoring of the lead time constant and `max_cycle_len` function. These seem to be used exclusively inside the `noisediode` module within `astrokat`. Please advise if external user scripts expect them at the top level of the package.

Improve the error handling when the lead time is too short. Check if we ran out of time and skip the rest of the noise diodes while reporting the error. We have other options here too:
   - Crash.
   - Retry the operation with a longer lead time based on how long it took thus far.

Crashing is a bit harsh for the extremely common scenario of just switching the noise diodes off. If the lead time is still too short, the operator could go and switch off noise diodes manually and the `astrokat-observe` script would then complain but continue. The alternative would involve either another hotfix or YAML hacking.

Refuse to handle an invalid timestamp to avoid crashes. There are options here too:
  - Don't bother reporting the average timestamp at the end of `_set_dig_nd_`. If you have `cycle=True`, the average won't make that much sense, and for the simulator it will be the final timestamp instead of the average anyway (i.e. it's a mess).
  - Don't return the timestamp from noise diode functions since they are ignored by the highest-level functionality. This way you can avoid spreading an invalid or hard-to-interpret timestamp.

### DIFF
--- a/astrokat/__init__.py
+++ b/astrokat/__init__.py
@@ -2,16 +2,6 @@
 from __future__ import division
 from __future__ import absolute_import
 
-# Constants and defaults
-_DEFAULT_LEAD_TIME = 5.0  # lead time [sec]
-def max_cycle_len(band):
-    if band.lower() == 'u':
-        return 31.  # buffer len [sec]
-    else:
-        # default is L-band
-        return 20.  # buffer len [sec]
-# Constants and defaults
-
 from .__main__ import cli
 
 from . import noisediode
--- a/astrokat/__init__.py
+++ b/astrokat/__init__.py
@@ -3,7 +3,7 @@ from __future__ import division
 from __future__ import absolute_import
 
 # Constants and defaults
-_DEFAULT_LEAD_TIME = 3.0  # lead time [sec]
+_DEFAULT_LEAD_TIME = 5.0  # lead time [sec]
 def max_cycle_len(band):
     if band.lower() == 'u':
         return 31.  # buffer len [sec]
--- a/astrokat/noisediode.py
+++ b/astrokat/noisediode.py
@@ -2,8 +2,10 @@
 from __future__ import division
 from __future__ import absolute_import
 
-import numpy as np
 import time
+
+import katpoint
+import numpy as np
 
 try:
     from katcorelib import user_logger
@@ -92,11 +94,11 @@ def _set_dig_nd_(kat,
         # The digitiser master controller takes about 15-50 ms per request,
         # so start panicking just before the deadline.
         if time.time() > timestamp - 0.02:
-            user_logger.error('Requested noise diode timestamp %s will probably '
+            user_logger.error('Requested noise diode timestamp %sZ will probably '
                               'be in the past - please increase lead time',
-                              time.ctime(timestamp))
-            skipped = ', '.join(nd_antennas[len(replies):])
-            user_logger.error('Skipped setting noise diodes on: %s', skipped)
+                              katpoint.Timestamp(timestamp))
+            skipped = ','.join(nd_antennas[len(replies):])
+            user_logger.error('Skipped setting these noise diodes: %s', skipped)
             break
         replies[ant] = ped.req.dig_noise_source(timestamp,
                                                 on_fraction,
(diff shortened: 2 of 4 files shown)


ajoubertza (approved):
Nice improvements.  Thanks @ludwigschwardt
   
Would be nice to hear from @rubyvanrooyen if the increased lead time could result in any other issues.

ludwigschwardt:
I've added a small commit to improve the logger output some more.

ludwigschwardt:
Hi, I've made another small commit to improve the warning message if a noise diode request fails.

@ajoubertza, is this still on track for tomorrow?
@rubyvanrooyen, any comments appreciated!

ajoubertza (approved):
The improvements to the logging look good.

rubyvanrooyen:
`For now, increase the lead time from 3.0 to 5.0 seconds`
Increasing the default lead time should not affect any of the standard observations.
I would propose erring on the side of caution and rather sticking to the longer lead times by default

Those users who desire a shorter lead time can specify the desired lead time through the YAML observation file

I updated the wiki doc to reflect this update to the ND lead time default setting
https://github.com/ska-sa/astrokat/wiki/Noise-diode-usage-during-observations

rubyvanrooyen:
`
Do some refactoring of the lead time constant and max_cycle_len function. These seem to be used exclusively inside the noisediode module within astrokat. Please advise if external user scripts expect them at the top level of the package.
`

`max_cycle_len` is a library parameter to check hardware setting limitations for the buffer lengths (values from Marc). It functionals as a sanity check only and I cannot see an immediate need to expose this at a higher level.

Tx for cleaning up my bad implementation of the parameter Lud

rubyvanrooyen:
`Crashing is a bit harsh for the extremely common scenario of just switching the noise diodes off.`

Switching the ND off can happen at:
* the start of the observation to ensure a default enter state (my current understanding from the astronomers is that here an error must rather abort the observation)
* the ND pattern or trigger event will switch the ND (obviously an error here will invalidate the requirements of the observation and the observation should be aborted)
* exiting the observation all NDs are switched off (it seems that this is where most of the time the error occur, so I made an option that will allow the exit to ignore errors here and continue with the rest of the commands)

I did a couple more tests with the pattern and trigger test YAMLs and added timestamp evaluations for the `on` and `pattern` functions. For the `off` function I included an option to ignore, which I also implement by default in the main observation script.

Please check my updates

rubyvanrooyen (commented):
In essence the I am happy for the merge to happen, but I do think the observation should fail if the ND settings will affect the expectation of the astronomer during observation.
Please just check my updates and make suggestions

ludwigschwardt:
Thanks Ruby, that looks good!

I'm happy to merge, especially since I'm on holiday next week... It would be good to get this version out on Wednesday, which will solve the current issues on site (i.e. get S-band and astrokat to co-exist).

When I'm back, we can explore some ideas to get the best of both worlds (strictness + flexibility :-D):

  - Make a special "noise diode off" routine (say `nd_reset`) that uses `'now'` timestamps. This will only be called at the beginning (and possibly end) of each script and therefore represents the main noise diode interaction. I realise that you removed `'now'` timestamps to get synchronised firings but I don't think the initial reset needs to be synchronised. This removes the troubles of guessing a sufficiently long lead time for 99% of noise diode firings.
  - Or, once we run out of time in `_set_dig_nd_`, instead of skipping the rest of the diodes we can guess a better lead time by comparing the number of antennas we've done so far against the total. Rerun `_set_dig_nd_` with the longer lead time, but then crash if we run out of time again. No need for infinite recursion :-D This will ensure a completely synchronised switch that is more robust against slow diodes.

So the error we are handling is really the inability of the script to set a long enough lead time, as opposed to something fundamentally wrong with the diodes. This is worth exploring because S-band diodes are coming... and they are dog slow thus far (maybe 30-40 seconds lead time for 64 dishes...). Until those diode controls are improved, it would be nice to have astrokat mostly working for commissioning to proceed.

rubyvanrooyen:
Tx Lud, I am merging the branch and closing the pull request

Your suggestions makes perfect sense, but I agree we need to get this fix out to site to at least control the current failures.
I captured you comment in issue #97 so it does not disappear when I merge and close this pr

The current implementation has been good for a long time, but the new receivers and modes will inevitably require updates :-) 
It makes sense to start a devel branch for this, since I want to do some other updates as well

ajoubertza:
@mmphego Please make a v28 hotfix for this, and arrange for it to be deployed on Wednesday 28 Apr.

PR #98, opened 2021-04-26
Hotfix-v28: Increase noise diode lead time and handle errors better

mmphego:
OPS-1899: Increase noise diode lead time and handle errors better

See https://github.com/ska-sa/astrokat/pull/96#issue-617837738 for PR details


lanceWilliams (approved):
Looks ok

PR #99, opened 2021-04-28
OPS-1899: Fix KATCP request check for success

ludwigschwardt:
I inadvertently switched to the `katcorelib` KATCP client style of checking success (used in CaptureSession), not realising that this uses raw KATCP.

Is there a way to test this on a real katcorelib to know that all is well before deployment?


rubyvanrooyen (approved):
looks good to me

lanceWilliams (approved):
Tested this in a development environment. Replicated the error and confirmed that this patch solved it. The test obs completed without any other failures so I'm happy to try to deploy with this fix again when we can.

ludwigschwardt:
Excellent, thanks @lanceWilliams !

PR #100, opened 2021-05-04
Fix katcp request reply check for noisediode

lanceWilliams:
This PR adds Ludwig's patch in https://github.com/ska-sa/astrokat/pull/99 to the astrokat release branch. This corrects a bug in how the katcp reply from the noisediode request is checked.


tockards (approved):
LGTM

PR #103, opened 2021-06-17
Support pyephem 4.0 time parser

ajoubertza:
Remove timezone info from `start_time` after loading from YAML.  If times have a trailing "Z", then the string representations are like 2019-02-11 02:10:47+00:00.  ephem <= 3.7.7.1 was OK to parse that (and ignored the timezone suffix).  Since version 4.0.0.1, the ephem parser is stricter and complains about these times.

This problem wasn't seen in the astrokat unit tests for this repo, but was picked up in the CAM AQF tests.  CAM AQF use katcorelib, instead of the simulated session.  The difference in katcorelib is that the start time is converted via `katpoint.Timestamp`, instead of `astrokat.datetime2timestamp`.

Added a unit test to ensure the conversions between astrokat and katpoint are equivalent, and exercise the ephem converter.

See more details [here](https://github.com/ska-sa/astrokat/pull/102/files#r651602048).

### DIFF
--- a/astrokat/test/test_offline_observe.py
+++ b/astrokat/test/test_offline_observe.py
@@ -3,10 +3,17 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import unittest
+import katpoint
 
 from mock import patch
 
-from .testutils import LoggedTelescope, execute_observe_main
+from astrokat import utility
+from .testutils import (
+    LoggedTelescope,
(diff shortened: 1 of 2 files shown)


mmphego (approved):
LGTM

rubyvanrooyen (approved):
Yes please, LGTM

PR #104, opened 2021-06-25
Support katcorelib in drift scan tests

ajoubertza:
The observation output messages differ between astrokat's [simulate](https://github.com/ska-sa/astrokat/blob/203e5f55797212029cdc87bbbee86f71764f6aa2/astrokat/simulate.py#L231-L233) module and katcorelib's [mkat_session](https://github.com/ska-sa/katcorelib/blob/c4f8e083a8afa26322a099f65d026f08a2255342/katcorelib/mkat_session.py#L1089-L1116) module.  The tests were only considering the messages generated by the simulate module.  Now we also check for messages in the format that katcorelib creates them.

This is to get CAM's AQF test run of astrokat's unit tests to pass:
```
======================================================================
FAIL: Check (az, el) target from (ra, dec) for drift scan
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "/home/kat/svn/astrokat/astrokat/test/test_scans.py", line 34, in test_drift_scan_basic_sim
    result)
AssertionError: 'Slewed to Az: -172:57:37.1 El: 56:27:26.4 at azel (-173.0, 56.5) deg' not found in "2021-06-24 19:11:25Z - Observation start up\n2021-06-24 19:11:25Z - Running astrokat version - 0.1.dev739+master.203e5f5\n2021-06-24 19:11:25Z - Skipping instrument checks - 20210624-0150 not in approved_schedule\n2021-06-24 19:11:25Z - 
```

PR #105, opened 2021-07-28
Hotfix v28: update to master

ajoubertza:
This brings the v28 release branch in line with changes in master, so it is based on PRs already reviewed.  Specifically: PRs #95, #101, #102, #103, #104 

The significant functional changes are:
  - Major cleanup of fundamental observation script to ensure consistent definition of targets and observations
  - Removal of target construction from main observation script into stand alone targets module that also handles coordinate conversion
  - Allow YAML input to handle galactic and scans using horizontal coordinates with tracking by converting these targets to equatorial (ra, dec) when observation starts.
  - Update scan target definition to bring it in line with normal tracking target specification. All targets will now require both coordinate and duration specification, no more special case scan duration implementation with slightly different meanings.
  - Ensure scans module works with the noise diode implementations, including lead time.
  - Update example YAML template to reflect updated implementation
  - Improvements to user scripts, including Google Colab notebooks (not directly used for MeerKAT observations) 

@rubyvanrooyen FYI


bngcebetsha (approved):
LGTM

PR #106, opened 2021-08-04
Add "reverse scan" over RADEC area (intensity mapping)

spassmoor:
Here is the pull request for the HI intensity mapping function.  "Reverse Scan"  you give coordinates of an area on the sky that you want to scan and the telescopes do a drift like scan over it.  Scanning in azimuth and  maintaining constant elevation while the source rises or sets.

The code reads in a set of coordinates from the scan parameters e.g. "radec_p1"  and works out a horizon that they will rise or set though. Once that horizon line is calculated the azimuth extent of these sources as they rise through that horizon line is determined and the furthest extent in azimuth. This will be the distance that the telescopes scan in azimuth . This distance  is used to  work out what the duration of the scan from the scan speed parameter. Note that this speed is scaled by 1/cos(el) to account for the projection effects.  which means very slow scan speeds at high elevations as a 1 deg scan at 80 deg elevation would have a 5.7 deg azimuth movement.  The scans move back and forth at a constant elevation until the lowest/highest  coordinates from the scan parameters has risen/set through that elevation. 

I have had issues with time zones and I hope that I have ensured that all the time calculations only use GMT/UTC


spassmoor:
I will quickly fix the PEP-8 Stuff

ajoubertza (commented):
Partial review, but fixed these as group effort

spassmoor:
I have updated the tests to have a second test that calls _get_scan_extents directly. I have also removed a few of the more useless tests.

astrokat/scans.py
```diff
@@ -174,18 +335,21 @@ def return_scan(session, target, nd_period=None, lead_time=None, **kwargs):
...
     user_logger.info("Reverse scan over target")
-    target_visible += reversescan(session,
-                                  target,
-                                  nd_period=nd_period,
-                                  lead_time=lead_time,
-                                  **kwargs)
+    returnscan = dict(kwargs)
+    returnscan["start"] = kwargs["end"]
+    returnscan["end"] = kwargs["start"]
+    target_visible += scan(session,
```
ajoubertza, line 348: Here is another `+=` used with booleans.   Should we be using `&=` instead? 

```python
In [1]: x = True

In [2]: x &= True

In [3]: x
Out[3]: True

In [4]: x &= False

In [5]: x
Out[5]: False

In [6]: x += True

In [7]: x
Out[7]: 1

In [8]: x += False

In [9]: x
Out[9]: 1

In [10]: x += True

In [11]: x
Out[11]: 2
```

astrokat/scans.py
```diff
@@ -146,15 +214,108 @@ def reversescan(session, target, nd_period=None, lead_time=None, **kwargs):
...
+
+    scan_start = np.degrees(az_min - (az_min + az_max) / 2.)
+    scan_end = np.degrees(az_max - (az_min + az_max) / 2.)
+
+    scanargs = {}
+    if "projection" in kwargs:
+        scanargs["projection"] = kwargs["projection"]
+
+    # take into account projection effects of the sky and convert to degrees per second
+    # 5 arcmin/s should translate to 5/cos(el)
```
ajoubertza, line 290: These comments on these two lines are a little confusing for me. 
The `5/cos(el)` takes the projection into account, but not the conversion from arcmin to degrees.  

Maybe:

```python
    # take into account projection effects of the sky and convert to degrees per second
    # E.g., 5 arcmin/s should translate to 5/60/cos(el) deg/s
```

astrokat/test/test_scans.py
```diff
@@ -49,6 +51,53 @@ def test_scan_basic_sim(self):
...
+            katpoint.Target("t3, radec, 06:44:00.0, -35:36:00.0", antenna=antenna),
+            katpoint.Target("t4, radec, 06:44:00.0, -25:42:00.0", antenna=antenna),
+        ]
+        el, az_min, az_max, t_start, t_end = scans._get_scan_area_extents(target_list,
+                                                                          antenna,
+                                                                          test_date)
+        return el, az_min, az_max, t_start, t_end
+
+    def test_reverse_scan_basic_sim(self):
+        """Test of drift scan over an area in the sky."""
```
ajoubertza, line 89:
```suggestion
        """Test of reverse scan over an area in the sky."""
```

astrokat/test/test_scans/reverse-scan-test.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name=PictorA_l0.5, radec=5:16:57.6653 -45:46:35.7098, tags=target, type=track, duration=70.0
+
+
+
+
+#
+#J1939-6342 | 1934-638, radec bpcal, 19:39:25.0264, -63:42:45.625, (408. 8640. -30.7667 26.4908 -7.0977 0.605334)
+#  J0519-4546 | PicA, radec bpcal fluxcal, 05:19:49.69, -45:46:44.5, (200. 4000. 3.513 -0.303 -0.074)
+#on_frac: 0.09
+#
```
ajoubertza, line 50: Can we delete all these lines?

obs_plans/reversescan.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - LST:  00:30-23:30
+    target_list:
+      - name=PictorA_u0.5, radec=5:19:49.721 -45:16:43.7801, tags=target, type=track, duration=70.0
+      - name=PictorA, radec=5:19:49.721 -45:46:43.7801, tags=target, type=track, duration=70.0
+      - name=PictorA_r0.5, radec=5:22:41.7767 -45:46:35.7098, tags=target, type=track, duration=70.0
+      - name=PictorA, radec=5:19:49.721 -45:46:43.7801, tags=target, type=track, duration=70.0
+      - name=PictorA_d0.5, radec=5:19:49.721 -46:16:43.7801, tags=target, type=track, duration=70.0
+      - name=PictorA, radec=5:19:49.721 -45:46:43.7801, tags=target, type=track, duration=70.0
+      - name=PictorA_l0.5, radec=5:16:57.6653 -45:46:35.7098, tags=target, type=track, duration=70.0
+      - name=scan_azel_with_nd_trigger, radec=23:00:00.0 -30:39:00, duration=120, tags=target, type=reversescan
```
ajoubertza, line 31: We don't need the duration for this one either.
```suggestion
      - name=scan_azel_with_nd_trigger, radec=23:00:00.0 -30:39:00, duration=0, tags=target, type=reversescan
```

obs_plans/reversescan.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name=PictorA_l0.5, radec=5:16:57.6653 -45:46:35.7098, tags=target, type=track, duration=70.0
+
+
+
+
+#
+#J1939-6342 | 1934-638, radec bpcal, 19:39:25.0264, -63:42:45.625, (408. 8640. -30.7667 26.4908 -7.0977 0.605334)
+#  J0519-4546 | PicA, radec bpcal fluxcal, 05:19:49.69, -45:46:44.5, (200. 4000. 3.513 -0.303 -0.074)
+#on_frac: 0.09
+#
```
ajoubertza, line 47: Please delete unnecessary lines.

ajoubertza (changes requested):
@spassmoor The new tests pass under CAM AQF now.  I've made a few more comments on minor things.  After that, I think this is done.

astrokat/scans.py
spassmoor: Done

astrokat/scans.py
spassmoor: Done

spassmoor:
I have made the requested changes.

ajoubertza (approved):
Looks good 👍

PR #110, opened 2021-08-26
Observation of noncelestial targets: planets and helio orbital comets

rubyvanrooyen:
Large new development to extension the observation framework to enable observation of planets as PyEphem special targets through katpoint.
As well as comets given XEphem orbital parameters

Update helper script to ensure all observation functionality will work for new targets

Add target definitions for all in the appropriate test directory for usage examples
Run all tests to ensure existing code is not broken by update

Update wiki pages to include new target definitions

PR #111, opened 2021-09-20
Hotfix V30: Update astrokat to support non-celestial targets

mamkhari:
This PR does the following:

- Large new development to extension the observation framework to enable observation of planets as PyEphem special targets through katpoint. As well as comets given XEphem orbital parameters
- Update helper script to ensure all observation functionality will work for new targets
- Add target definitions for all in the appropriate test directory for usage examples
- Run all tests to ensure existing code is not broken by the update
- Update wiki pages to include new target definitions

JIRA: [MT-2603](https://skaafrica.atlassian.net/browse/MT-2603)


ajoubertza (changes requested):
This looks OK @mamkhari.  However, since MKAT AQF tests aren't passing with the latest master branch that needs to be fixed first.  This PR needs to wait for the same fix.

mamkhari:
> This looks OK @mamkhari. However, since MKAT AQF tests aren't passing with the latest master branch that needs to be fixed first. This PR needs to wait for the same fix.

Oh okay, should I close this PR for now?

ajoubertza:
> Oh okay, should I close this PR for now?

No, don't close this one.  We will just add another commit to it later, to bring in the fix.

bngcebetsha:
> > Oh okay, should I close this PR for now?
> 
> No, don't close this one. We will just add another commit to it later, to bring in the fix.

@ajoubertza please re-review

astrokat/test/test_offline_observe.py
```diff
@@ -95,11 +98,13 @@ def assert_started_target_scan(
                 duration, target_string
             )
         )
+        sim_coordinates_string = re.findall(CORELIB_RADEC_REGEX, sim_radec)[0]
+        corelib_coordinates_string = re.findall(CORELIB_RADEC_REGEX, corelib_radec)[0]
```
ajoubertza, line 102: Sorry, but this doesn't look useful at all.  The original string is being extracted using a regex.

```python
In [1]: import re

In [2]: CORELIB_RADEC_REGEX = r"\d\d\:\d\d\:\d\d\.\d+\s-?\d+:\d\d\:\d\d\.\d+"

In [3]: corelib_radec = "15:59:15.06 8:50:58.7"

In [4]: re.findall(CORELIB_RADEC_REGEX, corelib_radec)[0]
Out[4]: '15:59:15.06 8:50:58.7'
```

The idea of the regex is to use it when searching the logs.  Rather pass the regex pattern to `assert_started_target_scan` function, so that it can apply it.  The regex pattern I gave @mamkhari was much more specific.

Here's an example.  I've used regex for both sim and corelib coordinates to make the code more consistent.

```python
    def test_targets_sim(self):
...
        self.assert_started_target_scan(
            "target4_azel",
            duration=10.0,
            az=10.0,
            el=50.0,
            sim_radec_regex=r"16:00:05.19 8:50:57.6",
            corelib_radec_regex=r"15:59:[0-5]\d\.\d+ 8:50:[0-5]\d\.\d+,
            logs=result
        )

...

    def assert_started_target_scan(
        self, target_string, duration, az, el, sim_radec_regex, corelib_radec_regex, logs
    ):
        ...
        simulate_radec_message_regex = r"{}, tags=radec target, {}".format(
            target_string, sim_radec_regex
        )
        katcorelib_radec_message_regex = r"{}, tags=radec target, {}".format(
            target_string, corelib_radec_regex
        )
        simulated_messages_found = (
            simulate_message in logs
            and re.search(simulate_radec_message_regex, logs, re.MULTILINE) is not None
        )
        katcorelib_messages_found = (
            katcorelib_message in logs
            and re.search(katcorelib_radec_message_regex, logs, re.MULTILINE) is not None
        )
```

astrokat/observe_main.py
```diff
@@ -38,14 +38,40 @@
...
+
+
+def _get_radec_from_azel(observer, orig_target_str, timestamp):
+    """Utility function to recalculate (ra, dec) from (az, el)"""
+    location = targets.observer_as_earth_location(observer)
+    az_deg, el_deg = np.array(orig_target_str.split(), dtype=float)
+    [ra_hms,
+     dec_dms] = targets.altaz_to_radec(az_deg, el_deg,
+                                       location, timestamp,
+                                       as_string=True)
```
mmphego, line 68: How'd the linter behave towards this?

astrokat/observe_main.py
```diff
@@ -365,134 +403,158 @@ def run_observation(opts, kat):
...
         # Target observation loop
-        with start_session(kat.array, **vars(opts)) as session:
-            session.standard_setup(**vars(opts))
+        observation_timer = time.time()
+        for obs_cntr, observation_cycle in enumerate(obs_plan_params["observation_loop"]):
+            if nr_obs_loops > 1:
+                user_logger.info("Observation loop {} of {}."
+                                 .format(obs_cntr + 1, nr_obs_loops))
+                user_logger.info("Loop LST range {}."
+                                 .format(observation_cycle["LST"]))
```
mmphego, line 431: Instead of using `.format()` str manipulation logic within the logger, you could use the recommend str formatting for Python logs.

Suggestion:

```python
user_logger.info('Post with id %s not found', post_id)
```

The above is more efficient and less expensive.

Below I added a screenshot, highlighting the cost per log string formatting. Notice the native `%s` string formatting is 24ns faster than concatenating/adding 2/more string together and 124ns faster than using `.format()` method. :wink:



Some recommended reading:

- [Stackoverflow: Python - Logger string formatting](https://stackoverflow.com/a/13375482)
- [Python Logging Best Practices](https://coralogix.com/blog/python-logging-best-practices-tips/)

astrokat/observe_main.py
```diff
@@ -365,134 +403,158 @@ def run_observation(opts, kat):
...
-                        "{}-{}".format(
-                            ephem.hours(str(start_lst)), ephem.hours(str(end_lst))
-                        )
+            user_logger.info(
+                "Observation targets are [{}]".format(
+                    ", ".join(
+                        [
+                            repr(target.name)
+                            for target in catalogue.filter(not_cals_filter_list)
+                        ]
```
mmphego, line 554: Perhaps, simplify this by assigning this logic to a var and then calling the var here... this is a bit harder to read.

scripts/astrokat-catalogue2obsfile.py
```diff
@@ -329,7 +362,8 @@ def write_yaml(self, header=None, configuration=None, outfile="obs_config.yaml")
             }
             instrument = vars(argparse.Namespace(**group_dict))
             break
-    for key in instrument.keys():
+    instrument_keys = list(instrument.keys())
+    for key in instrument_keys:
```
mmphego, line 366: You could've gotten away without changing this logic.

scripts/astrokat-targets.py
```diff
@@ -491,12 +507,16 @@ def table_line(datetime,
     [rise_time,
      set_time] = observatory.target_rise_and_set_times(target.body,
                                                        lst=lst)
-    if not lst:
+
+    if type(rise_time) is ephem.Angle:
```
mmphego, line 511: When testing for the type, you should instead use [isinstance(tmpDict[key], list)](https://docs.python.org/3.4/library/functions.html#isinstance) when testing the type of something, apparently, it is a more Pythonic way of checking the type :smile: 
```suggestion
    if isinstance(ephem.Angle, rise_time):
```

scripts/astrokat-targets.py
```diff
@@ -918,12 +962,24 @@ def main(creation_time,
...
-        target = [tgt.strip() for tgt in target]
-        target = ", ".join(
-            map(str, [target[0], "radec target", target[1], target[2]])
-        )
-        cal_targets = [katpoint.Target(target)]
+        if len(target) > 2:
+            # input target from command line
+            target = [tgt.strip() for tgt in target]
+            target = ", ".join(
+                map(str, [target[0], "radec target", target[1], target[2]])
```
mmphego, line 969: :+1: this is cool, I like this map thing.

scripts/astrokat-targets.py
```diff
@@ -1060,11 +1116,28 @@ def main(creation_time,
...
                    'pi_name': args.pi,
                    'pi_contact': args.contact}
+    if args.target is not None:
+        target = args.target
+    elif args.body is not None:
+        target = [args.body]
+    elif args.xephem is not None:
+        target = args.xephem
+    else:
+        target = None
```
mmphego, line 1126: That new fancy Python switch statement would be useful here. Pity, it's only on Py3.10

ajoubertza:
@mmphego Thanks for the comments, but those would have to be applied to the master branch (this work was already merged in PR #110).

ajoubertza (approved):
LGTM.  The AQF tests have passed after #112.

PR #112, opened 2021-10-12
Fix AQF tests for test_targets_sim

bngcebetsha:
For details please see - https://github.com/ska-sa/astrokat/pull/111

The changes need to be first merged into master.


mmphego (approved):
FYI: https://stackoverflow.com/a/26318298

PR #113, opened 2022-04-25
Update strokat to use new set target functionality and remove time consuming calls to fbfuse

bngcebetsha:
mkat_session has been updated to perform a "slow only" feature which is set in the `set_target` and `track` methods. This is useful for when slewing to the first target and we do not spend time doing setting for `fbf` - see https://github.com/ska-sa/katcorelib/pull/369 for more details.

In this PR, we update astrokat to use the "slew only" functionality when running observation. We also update the unittests
JIRA: [MT-2952](https://skaafrica.atlassian.net/browse/MT-2952)

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -69,6 +69,8 @@ def _get_radec_from_azel(observer, orig_target_str, timestamp):
     return ra_hms, dec_dms
 
 # -- Utility functions --
+
+
 def initial_slew(session, target_info):
     """Simple way to get telescope to slew to target
     Parameters
@@ -80,13 +82,12 @@ def initial_slew(session, target_info):
     katpt_tgt = target_info["target"]
 
     user_logger.info("Slewing to target {}".format(target_name))
-    # modify set_target to use the new target
-    session.set_target(katpt_tgt)
-    session.activity('slew')
+    session.set_target(katpt_tgt, slew_only=True)
 
     if session.kat.dry_run:
         # Apply average slew time
         session._slew_to(katpt_tgt)
+        #call track() loka time session this will take in the slew_only 
     else:
         # Start moving each antenna to the target
         session.ants.req.mode('POINT')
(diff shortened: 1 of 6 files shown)


astrokat/observe_main.py
```diff
@@ -71,14 +71,49 @@ def _get_radec_from_azel(observer, orig_target_str, timestamp):
...
+        session._slew_to(katpt_tgt)
+    else:
+        # Start moving each antenna to the target
+        session.ants.req.mode('POINT')
 
+    # Wait until a quorum is in position (with timeout)
+    success = session.wait(session.ants,
+                           'lock',
+                           True,
+                           timeout=300,
```
mmphego, line 101: Perhaps, pass the timeout value as a variable.

astrokat/observe_main.py
```diff
@@ -599,8 +634,8 @@ def run_observation(opts, kat):
             )
 
             # Go to first target before starting capture
-            user_logger.info("Slewing to first target")
-            observe(session, ref_antenna, obs_targets[0], slewonly=True)
+            initial_slew(session, obs_targets[0])
+            # observe(session, obs_targets[0], slew_only=True)
```
mmphego, line 638: Remove dead code.

astrokat/simulate.py
```diff
@@ -197,7 +197,7 @@ def __nonzero__(self):
 
     def __iter__(self):
         yield self
-        raise StopIteration
+        return
```
mmphego, line 200: Why return instead of raising a StopIteration?

astrokat/simulate.py
```diff
@@ -216,7 +216,16 @@ def capture_init(self):
             user_logger.info('INIT')
             self.capture_initialised = True
 
-    def track(self, target, duration=0, announce=False):
+    def wait(self, *args, **kwargs):
```
mmphego, line 219: Why you passing *args and **kwargs, if they are not being used?

lanceWilliams (approved):
Looks ok. I think during the next few weeks we should organise some kind of test with Operations, in particular Sarah because I would like to see if there is a measurable difference in performance or if a special yaml file is required to exploit this change.

PR #115, opened 2022-11-22
Implement reference pointing in astrokat

bngcebetsha:
We recently added a function that will perform reference pointing scans in katcorelib. In this PR we modify astrokat to enable that same functionality for astrokat observe script.

Tasklog - http://10.8.67.120:8081/tailtask/20221129-0011/progress
Will wait for the right LST to start observation to see outcomes.

JIRA: [MT-3213](https://skaafrica.atlassian.net/browse/MT-3213)

### DIFF
--- a/astrokat/test/test_nd_timings.py
+++ b/astrokat/test/test_nd_timings.py
@@ -64,7 +64,7 @@ class TestAstrokatYAML(unittest.TestCase):
 
         result = LoggedTelescope.user_logger_stream.getvalue()
         self.assertIn("No ND for target", result)
-        self.assertIn("noise-diode off at 1573714914.0", result)
+        self.assertIn("noise-diode off at 1573715222.0", result)
         self.assertIn("Restoring ND pattern", result)
         self.assertIn("noise diode pattern every 0.1 sec, with 0.05 sec on",
                       result)
--- a/astrokat/test/test_scans.py
+++ b/astrokat/test/test_scans.py
@@ -52,6 +52,14 @@ class TestAstrokatYAML(unittest.TestCase):
         self.assertIn("Initialising Scan target scan_1934-638 for 30.0 sec", result)
         self.assertIn("scan_1934-638 observed for 60.0 sec", result)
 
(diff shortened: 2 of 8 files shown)


astrokat/test/test_scans.py
```diff
@@ -52,6 +52,14 @@ def test_scan_basic_sim(self):
         self.assertIn("Initialising Scan target scan_1934-638 for 30.0 sec", result)
         self.assertIn("scan_1934-638 observed for 60.0 sec", result)
 
+    def test_reference_pointing_scan_basic_sim(self):
+        """Not much to do: check scan initiate log msg"""
+        execute_observe_main("test_scans/reference-pointing-scan-test.yaml")
+        # get result and make sure everything ran properly
+        result = LoggedTelescope.user_logger_stream.getvalue()
+        self.assertIn("Initialising Scan target scan_1934-638 for 30.0 sec", result)
```
lanceWilliams, line 60: yaml configuration does not seem to align with the scan durations that lines 60 and 61 are looking for. I wouldn't expect this to pass as is.

lanceWilliams (changes requested):
Unittests seem to be failing on Jenkins due to a keyError. Please confirm.

astrokat/test/test_nd_timings.py
```diff
@@ -64,7 +64,7 @@ def test_nd_pattern_plus_off(self):
 
         result = LoggedTelescope.user_logger_stream.getvalue()
         self.assertIn("No ND for target", result)
-        self.assertIn("noise-diode off at 1573714914.0", result)
+        self.assertIn("noise-diode off at 1573715222.0", result)
```
lanceWilliams, line 67: Making this as a general comment: originally, I mentioned the scan duration relating to the reference scan because the log line the test seemed to be looking for had a different value to what the scan was configured for in its yaml file. I see in the newer commits that there are now changes to other tests as well which I assume are not reference scan tests or use the reference scan yaml. Were these tests failing before? There seems to be some discrepancy in the log line checks.

PR #116, opened 2023-02-04
Improve reference pointing functionality

bngcebetsha:
The function call to `reference_pointing_scan` was placed among the "scans" and led to a confusion in how many tracks of the target during calls `reversescan`. We now move the check if the observation is "reference_pointing" to earlier in our logic.

Reference pointing gives the expected telstateError result - http://10.8.67.104:8081/tailtask/20230206-0001/progress

#### Before fix
[MASTER.txt](https://github.com/ska-sa/astrokat/files/10667766/MASTER.txt)

#### After fix
[after_fix.txt](https://github.com/ska-sa/astrokat/files/10667761/after_fix.txt)

JIRA: [MT-3289](https://skaafrica.atlassian.net/browse/MT-3289)

[MT-3289]: https://skaafrica.atlassian.net/browse/MT-3289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -145,7 +145,6 @@ def observe(session, ref_antenna, target_info, **kwargs):
                                           lead_time=nd_lead)
     elif "reference_pointing_scan" in obs_type:
         scan_func = scans.reference_pointing_scan
-        obs_type = "reference_pointing_scan"
         target_visible = scan_func(session,
                                    target,
                                    nd_period=nd_period,
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -733,10 +733,10 @@ def run_observation(opts, kat):
                             "observed {}".format(target["last_observed"])
                         )
 
-                        targets_visible += observe(session,
+                        targets_visible += bool(observe(session,
                                                    ref_antenna,
                                                    target,
-                                                   **obs_plan_params)
+                                                   **obs_plan_params))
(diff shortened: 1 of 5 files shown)


spassmoor (approved):
This looks ok .

lanceWilliams:
@bngcebetsha could you explain why casting the function output to bool fixes the testing issue?

bngcebetsha:
> @bngcebetsha could you explain why casting the function output to bool fixes the testing issue?

hi @lanceWilliams ,....the `targets_visible` variable is of type `bool` and for some slews in the test, `targets_visible` is `None` - which causes failures "cannot use + operator on NoneType" sort of error.

ludwigschwardt:
Hi @bngcebetsha, why is `targets_visible` `None`? Let's see if we can rather fix that instead, instead of casting `None` to `False`...

lanceWilliams (approved):
The return type was missing from the simulated session reference pointing function. I've added it and the tests now work without needing to cast the output. @bngcebetsha please review from your side and repeat your test exercises (possibly ping Sean again as well).

ludwigschwardt (approved):
Excellent!

spassmoor:
Yes the changes look Good

PR #117, opened 2023-03-07
Hotfix V30: Improve reference pointing functionality

bngcebetsha:
The function call to `reference_pointing_scan` was placed among the "scans" and led to a confusion in how many tracks of the target during calls `reversescan`. We now move the check if the observation is "reference_pointing" to earlier in our logic.

Reference pointing gives the expected telstateError result - http://10.8.67.104:8081/tailtask/20230206-0001/progress

JIRA: [MT-3289](https://skaafrica.atlassian.net/browse/MT-3289)


bngcebetsha:
thanks for picking the missing code @lanceWilliams . This looks complete now

gtrymore (approved):
LGTM

PR #118, opened 2023-08-23
Issue noisediode deactivation to happen now

bngcebetsha:
As reported in [OPS-2080](https://skaafrica.atlassian.net/browse/OPS-2080), the observation script switches the noisediodes off at the end of observations. There have been issues on site where lost connection to a receptor results in failure to switch the noisediodes off on all other antennas on the array, which reports a failure. Here we attempt to mitigate that by using the current time as the time to reset the noisediode.

## PLEASE NOTE
We still need to test this, so please take this as a draft PR.

JIRA: [MT-3844](https://skaafrica.atlassian.net/browse/MT-3844)

[OPS-2080]: https://skaafrica.atlassian.net/browse/OPS-2080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MT-3844]: https://skaafrica.atlassian.net/browse/MT-3844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/astrokat/noisediode.py
+++ b/astrokat/noisediode.py
@@ -194,13 +194,8 @@ def nd_reset(kat,
     timestamp : float
         Linux timestamp reported by digitiser
     """
-    antennas = sorted(ant.name for ant in kat.ants)
     user_logger.info('Resetting all noise diodes')
-    # TODO - bngcebetsha: do we want to do this for ALL antennas or a targeted group such
-    # as is done in `_set_dig_nd_()`?
-    for ant in antennas:
-        ped = getattr(kat, ant)
-        ped.req.dig_noise_source(timestamp, switch)
+    kat.ants.req.dig_noise_source(timestamp, switch)
 
 
 def _switch_on_off_(kat,
--- a/astrokat/noisediode.py
+++ b/astrokat/noisediode.py
@@ -195,7 +195,7 @@ def nd_reset(kat,
         Linux timestamp reported by digitiser
     """
     antennas = sorted(ant.name for ant in kat.ants)
-    user_logger.info('Switching off all noise diodes')
+    user_logger.info('Resetting all noise diodes')
     # TODO - bngcebetsha: do we want to do this for ALL antennas or a targeted group such
     # as is done in `_set_dig_nd_()`?
     for ant in antennas:
--- a/astrokat/noisediode.py
+++ b/astrokat/noisediode.py
@@ -103,6 +103,10 @@ def _set_dig_nd_(kat,
         replies[ant] = ped.req.dig_noise_source(timestamp,
                                                 on_fraction,
                                                 cycle_length)
+        # TODO - bngcebetsha: update the switcing off of noise diodes to happen 'now' in
+        # order to avoid attempting to set them when the `timestamp` is in the past.
+        # user_logger.info('Switching off all noise diodes')
+        # replies[ant] = ped.req.dig_noise_source('now', 0)
         if kat.dry_run:
             msg = ('Dry-run: Set noise diode for antenna {} at '
                    'timestamp {}'.format(ant, timestamp))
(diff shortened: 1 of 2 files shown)


astrokat/observe_main.py
```diff
@@ -312,7 +312,8 @@ def __exit__(self, type, value, traceback):
         # TODO: Return correlator settings to entry values
         # switch noise-source pattern off (ensure this after each observation)
         # if NaN returned at off command, allow to continue
-        noisediode.off(self.array, allow_ts_err=True)
+        noisediode.nd_reset(self.array, time.time())
```
ludwigschwardt, line 315: There is a special timestamp called `"now"` that simplifies this a lot...

astrokat/noisediode.py
```diff
@@ -175,6 +175,34 @@ def _nd_log_msg_(ant,
...
+    Returns
+    -------
+    timestamp : float
+        Linux timestamp reported by digitiser
+    """
+    antennas = sorted(ant.name for ant in kat.ants)
+    user_logger.info('Resetting all noise diodes')
+    # TODO - bngcebetsha: do we want to do this for ALL antennas or a targeted group such
+    # as is done in `_set_dig_nd_()`?
+    for ant in antennas:
```
ludwigschwardt, line 201: There is no need to loop over antennas... Use the `kat.ants` group instead. I.e. `kat.ants.req.dig_noise_source(timestamp, switch)`.

ludwigschwardt (changes requested):
I'm still going through this but I suspect there is a much simpler way to do this... See e.g. https://skaafrica.atlassian.net/browse/OPS-2790?focusedCommentId=164604

lanceWilliams:
@ludwigschwardt just following up on this: so would you be ok if we removed the noise diode off command in https://github.com/ska-sa/astrokat/blob/f7e29ce091100a6041c61a8f3974f7b6a2611adf/astrokat/observe_main.py#L301  (and rely on session to do it) and then retain the function Buntu has created to turn noisediodes off with the 'now' timestamp on the way out of the schedule block in the __exit__  function ?

astrokat/noisediode.py
```diff
@@ -175,6 +175,29 @@ def _nd_log_msg_(ant,
     return actual_time
 
 
+def nd_reset(kat,
+             timestamp,
```
ludwigschwardt, line 179: You say in the docstring that this is optional but there is no default value... I would just make the default value "now".

astrokat/noisediode.py
```diff
@@ -175,6 +175,29 @@ def _nd_log_msg_(ant,
...
+        Time since the epoch as a floating point number [sec]
+    switch: int, optional
+        off = 0 (default), on = 1
+
+    Returns
+    -------
+    timestamp : float
+        Linux timestamp reported by digitiser
+    """
+    user_logger.info('Resetting all noise diodes')
```
ludwigschwardt, line 197: I would probably report whether you are switching it on or off.

astrokat/noisediode.py
```diff
@@ -175,6 +175,29 @@ def _nd_log_msg_(ant,
...
+    kat : session kat container-like object
+        Container for accessing KATCP resources allocated to schedule block.
+    timestamp : float, optional
+        Time since the epoch as a floating point number [sec]
+    switch: int, optional
+        off = 0 (default), on = 1
+
+    Returns
+    -------
+    timestamp : float
```
ludwigschwardt, line 194: I don't think it returns anything.

ludwigschwardt (changes requested):

ludwigschwardt:
Lance, your suggestion makes sense. I don't switch them off on the way out, just on the way in.

ludwigschwardt (approved):

Issue #120, opened 2024-04-22
drift scans


mariogrs:
Is the drift scan observation type properly implemented? I tried it and astrokat-observe gave some weird output with negative elevation:

2024-04-22 15:02:48Z - Setting up telescope for observation
2024-04-22 15:02:48Z - Observation start up
2024-04-22 15:02:48Z - Running astrokat version - 0.0+unknown.202404221703
2024-04-22 15:02:48Z - Skipping instrument checks - approved_schedule does not exist
2024-04-22 15:03:12Z - Found 1 target(s): 0 from 0 catalogue(s), 0 from default catalogue and 1 as target string(s)
2024-04-22 18:00:00Z - Observation targets are ['3C273']
2024-04-22 18:00:00Z - Set noise diode period to multiple of correlator dump period: cycle length = 19.5 [sec]
2024-04-22 18:00:00Z - Request: Set noise diode pattern to activate at 1713808812.0 (includes 12.0 sec lead time)
2024-04-22 18:00:00Z - Antennas found in subarray, setting ND: m011,m022,m033,m044
2024-04-22 18:00:00Z - Repeat noise diode pattern every 19.5 sec, with 0.585 sec on and apply pattern to ['m011', 'm022', 'm033', 'm044']
2024-04-22 18:00:12Z - Report: Switch noise-diode pattern on at 1713808812.0
2024-04-22 18:00:12Z - Waiting for observation setup
2024-04-22 18:00:15Z - INIT
2024-04-22 18:00:15Z - Slewing to first target
2024-04-22 18:01:00Z - Slewed to 3C273 at azel (61.0, 36.3) deg
2024-04-22 18:01:00Z - Tracked 3C273 for 0 seconds
2024-04-22 18:01:00Z - Initialising Drift_scan target 3C273 for 400.0 sec
2024-04-22 18:01:00Z - Drift_scan observation for 400.0 sec
2024-04-22 18:02:34Z - Slewed to Az: -107:47:45.3 El: -31:05:16.1 at azel (-107.8, -31.1) deg
2024-04-22 18:09:14Z - Tracked Az: -107:47:45.3 El: -31:05:16.1 for 400 seconds
2024-04-22 18:09:14Z - Observation list completed - ending observation
2024-04-22 18:09:14Z - Observation loop statistics
2024-04-22 18:09:14Z - Single run through observation target list
2024-04-22 18:09:14Z - Total observation time 554.70 sec (9.25 min)
[...]

PR #121, opened 2024-04-30
Fix Timestamp issue on driftscan

spassmoor:
This bug was raised in ticket  #120 

Adding an float  to a ephem_date date object returns an float number of days from 1900 plus the float value. 
This when passed to a katpoint target object is interpreted as a unix epoch,  number of seconds from 1970.  
There are two bugs in the code : 
1. First the duration that is added is in seconds but when added to ephem_date is interperated as days. 
2. Second the result which is days is assumed to be unix epoch seconds. 
 
This is fixed by first converting the ephem_date object to a katpoint.Timestamp object then adding the float number of seconds. Then using  the to_ephem_date method of the katpoint.Timestamp to convert it back.

### DIFF
--- a/astrokat/scans.py
+++ b/astrokat/scans.py
@@ -32,7 +32,7 @@ def drift_pointing_offset(ref_antenna, target, duration=60.0):
 
     """
     obs_start_ts = ref_antenna.observer.date
-    transit_time = obs_start_ts + duration / 2.0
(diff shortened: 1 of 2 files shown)


lanceWilliams (approved):
Looks ok. I'll talk to System Team on Thursday about pulling to site.

PR #126, opened 2025-02-25
SPR1-3207: Add relevant pointing parameters to session options to enable adjust pointing 

Tasmiyah27:
Previously, `obs_type` was set to a normal '`scan`', this is corrected by replacing it with a '`reference_pointing_scan`'. Total duration of reference pointing scan is added to '`duration`' key in obs_type dictionary in kwargs which later gets passed to, and used, in the `reference_pointing_scan` function in astrokat/scans.py.
The adjust pointing option is added to session options which gets passed to `mkat_session`. Here `obs_plan_params["adjust_pointing"]` is a boolean which will enable pointing adjustment if True. Adjust pointing parameters is added to session options, ie. `max_age` and `max_dist` which defines the maximum age of the solutions and maximum distance of pointing calibrator away from target that can be used to adjust pointing.

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -166,7 +166,11 @@ def observe(session, ref_antenna, target_info, **kwargs):
             obs_type = "scan"
         elif "reference_pointing_scan" in obs_type:
             scan_func = scans.reference_pointing_scan
-            obs_type = "scan"
+            obs_type = "reference_pointing_scan"
+            if obs_type not in kwargs.keys():
+                kwargs[obs_type] = {'duration': duration}
+            else:
+                kwargs[obs_type]['duration'] = duration           
         elif "return_scan" in obs_type:
             scan_func = scans.return_scan
             obs_type = "scan"
@@ -390,6 +394,8 @@ def run_observation(opts, kat):
     obs_plan_params = opts.obs_plan_params
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
+    #TODO: rather pass obs_plan_params in observe.py in astrokat
+    #to save in telstate
 
     # set up duration periods for observation control
     obs_duration = -1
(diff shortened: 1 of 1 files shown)


astrokat/observe_main.py
```diff
@@ -166,7 +166,11 @@ def observe(session, ref_antenna, target_info, **kwargs):
             obs_type = "scan"
         elif "reference_pointing_scan" in obs_type:
             scan_func = scans.reference_pointing_scan
-            obs_type = "scan"
+            obs_type = "reference_pointing_scan"
+            if obs_type not in kwargs.keys():
```
ludwigschwardt, line 170: Replace `in kwargs.keys()` with just `in kwargs`. I see a few places that need help 😅

astrokat/observe_main.py
```diff
@@ -391,6 +391,9 @@ def run_observation(opts, kat):
     # remove observation specific instructions housed in YAML file
     del opts.obs_plan_params
 
+    # TODO: rather pass obs_plan_params in observe.py in astrokat
```
ludwigschwardt, line 394: I think remove this TODO until we have clarity on what to do.

astrokat/observe_main.py
```diff
@@ -413,6 +416,15 @@ def run_observation(opts, kat):
             description = opts.proposal_description
         session_opts["description"] = description
 
+    if "adjust_pointing" not in vars(opts):
+        session_opts = vars(opts)
+        adjust_pointing = obs_plan_params["adjust_pointing"]
```
ludwigschwardt, line 421: Would this key be at the top level of the YAML? I thought it was under the "reference_pointing_scan" section.
Also, I would make it `False` if that section is not there.

astrokat/observe_main.py
```diff
@@ -413,6 +416,15 @@ def run_observation(opts, kat):
...
+        max_age = obs_plan_params["max_age"]
```
ludwigschwardt, line 422: These two also come out of the "reference_pointing_scan" section.
I would make them optional, only setting them if they are there.

ludwigschwardt (changes requested):
Think about adding some tests as well...

astrokat/observe_main.py
Tasmiyah27: should I do this for all other instances in the file

astrokat/observe_main.py
Tasmiyah27: I believe we wanted it in the yaml as a True/ False value
Do you mean making it something like:
```
if "adjust_pointing" not in vars(opts):
        session_opts = vars(opts)
        adjust_pointing = obs_plan_params["adjust_pointing"]
        if adjust_pointing == True:
            max_age = obs_plan_params["max_age"]
            max_dist = obs_plan_params["max_dist"]
            session_opts["adjust_pointing"] = adjust_pointing
            session_opts["pointing_solution_max_age"] = max_age
            session_opts["pointing_solution_max_dist"] = max_dist
```

since adjust_pointing is already a boolean that would be False if not selected

PR #114, opened 2022-05-03
Hotfix V30: Update astrokat to use new set target functionality

bngcebetsha:
`mkat_session` has been updated to perform a "slow only" feature which is set in the `set_target` and `track` methods. This is useful for when slewing to the first target and we do not spend time doing setting for `fbf` - see https://github.com/ska-sa/katcorelib/pull/369 for more details.

In this PR, we update astrokat to use the "slew only" functionality when running observation.
JIRA: [MT-2952](https://skaafrica.atlassian.net/browse/MT-2952)

####  Testing before deployment
- To run the unittests for this branch, please follow this part of the README
`cd astrokat`
Python virtual environment setup
```
python3 -m venv env
source env/bin/activate
pip install nose
pip install mock
pip install -e .
```
Then head to the test directory and run the tests:
```
cd astrokat/test
`python3 -m nose .
```
For system level testing, install the branch
```
cd svn/astrokat
sudo pip install . -U --no-deps
```
Then run this schedule block
`ipython`
```python
import katuilib
configure_obs()
obs.sb.new(owner='AstroKAT')
obs.sb.type=katuilib.ScheduleBlockTypes.OBSERVATION
obs.sb.antenna_spec='available'
obs.sb.controlled_resources_spec='cbf,sdp'
obs.sb.description='nd-pattern-sim'
obs.sb.proposal_id='devel'
obs.sb.instruction_set="run-obs-script /home/kat/svn/astrokat/scripts/astrokat-observe.py --yaml /home/kat/svn/astrokat/astrokat/test/test_nd/nd-pattern-sim.yaml"
obs.sb.to_defined()
obs.sb.to_approved()
obs.sb.unload()
```

[MT-2952]: https://skaafrica.atlassian.net/browse/MT-2952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #129, opened 2025-09-22
SPP1-168: Set CBF target to middle (ra, dec) of each scan in OTF mapping

tamerakassie:
Created the  ra dec target to be set on CBF for delay tracking in reverse scan observations, we will create a default target dictionary where the `cbf` key will correspond to the ra dec katpoint target that is fixed and the scans of the antennas will run across at `time_radec` i.e ~ midpoint of a scan line, and the `ants` key will correspond to the `scan_azel_with_nd_trigger`. The interface to `CaptureSession.set_target`is refactored to accept a dictionary of targets instead of a single target, in `reversescan` we create a simple target dictionary which contain the following keys: `cbf`, `ants` with respective  `scan_azel_with_nd_trigger`for the area extent scan and `radec target` for delay tracking on CBF, these will be passed to `CaptureSession`.

To maintain the code flexibility and logic  we use an `instance` checks for the target dictionary to access the `scan_target` or `radec target`  when required by the reverse scan simulations, while not affecting the other scan simulations in  the other `astrokat` tests.

### DIFF
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -97,6 +97,20 @@ def observe(session, ref_antenna, target_info, **kwargs):
         target.body._dec = dec_dms
 
     # simple way to get telescope to slew to target
+    #
+    # NOTE: This small bit of code has had a troubled history :-)
+    # - started off as a basic 0-second track
+    # - #92: Replaced with fancy slew function to accommodate FBFUSE
+    # - #93: Reverted back to 0-second track
+    # - #113: Used new "slew_only" functionality in session.track
+    # - #129: Stripped that out again since #114 was never released to site
+    #
+    # If you want to restore "slew_only" functionality with the new
+    # target dict approach, do the following:
+    #
+    #   session.track({"ants": target}, duration=0.0, announce=False)
+    #
+    # But this might not help the FBFUSE's cause...
     if "slewonly" in kwargs:
         return session.track(target, duration=0.0, announce=False)
 
--- a/astrokat/observe_main.py
+++ b/astrokat/observe_main.py
@@ -98,7 +98,7 @@ def observe(session, ref_antenna, target_info, **kwargs):
 
     # simple way to get telescope to slew to target
     if "slewonly" in kwargs:
-        return session.track(target, duration=0.0, announce=False, slew_only=True)
+        return session.track(target, duration=0.0, announce=False)
 
     # set noise diode behaviour
     nd_setup = None
--- a/astrokat/scans.py
+++ b/astrokat/scans.py
@@ -111,11 +111,11 @@ def scan(session, target, nd_period=None, lead_time=None, **kwargs):
 
     """
     if isinstance(target, dict):
-        scan_target = target.get('ants')
-        radec_target = target.get('cbf')
+        ants_target = target.get('ants')
+        cbf_target = target.get('cbf')
     else:
-        scan_target = target
-        radec_target = target
+        ants_target = target
+        cbf_target = target
 
     # trigger noise diode if set
     trigger(session.kat, duration=nd_period, lead_time=lead_time)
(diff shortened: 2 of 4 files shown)


astrokat/scans.py
```diff
@@ -110,14 +110,23 @@ def scan(session, target, nd_period=None, lead_time=None, **kwargs):
         noisediode trigger lead time
 
     """
+    if isinstance(target, dict):
+        scan_target = target.get('ants')
+        radec_target = target.get('cbf')
```
lanceWilliams: Could we just call this the 'cbf_target' as opposed to the 'radec_target'? As far as I am aware the 'scan_target' is almost always a 'radec' target anyway.

astrokat/scans.py
```diff
@@ -331,8 +339,19 @@ def reversescan(session, target, nd_period=None, lead_time=None, **kwargs):
             scanargs["end"] = scan_start, 0.0
         user_logger.info("Azimuth scan extent [%.1f, %.1f]" %
                          (scanargs["start"][0], scanargs["end"][0]))
+
+        # Determine the cbf targets
+        # use scan_target to the astrometric (ra, dec)
```
lanceWilliams: This line and 345 read strangely. Are these typos?

astrokat/simulate.py
```diff
@@ -365,21 +369,29 @@ def _target_azel(self, target):
             The elevation co-ordinate of the target in degrees.
 
         """
-        az, el = target.azel(simobserver.date)
+        if isinstance(target, dict):
+            scan_target = target.get('ants')
```
lanceWilliams: I think this simulation function is shared between tracks and scans? Referencing 'scan_targets' here is perhaps misleading.

astrokat/simulate.py
```diff
@@ -365,21 +369,29 @@ def _target_azel(self, target):
...
+        else:
+            scan_target = target
+        az, el = scan_target.azel(simobserver.date)
         az = katpoint.rad2deg(az)
         el = katpoint.rad2deg(el)
         return az, el
 
     def _fake_slew_(self, target):
+        if isinstance(target, dict):
+            scan_target = target.get('ants')
```
lanceWilliams: As the with the comment above, I think this simulation function is shared between tracks and scans? Referencing 'scan_targets' here is perhaps misleading.

astrokat/scans.py
ludwigschwardt: It is a bit unclear, yes. I would just say something like:
```
# CBF target is (ra, dec) position of reference antenna in middle of next scan
```

astrokat/observe_main.py
```diff
@@ -98,7 +98,7 @@ def observe(session, ref_antenna, target_info, **kwargs):
 
     # simple way to get telescope to slew to target
     if "slewonly" in kwargs:
-        return session.track(target, duration=0.0, announce=False, slew_only=True)
+        return session.track(target, duration=0.0, announce=False)
```
lanceWilliams, line 101: With the proposed changes in katcorelib it seems like the 'slewonly' kwarg wouldn't do anything unless it altered the structure of the target mapping being passed to 'track'?

ludwigschwardt: It does exactly nothing, which is why we need to decide what to do about the discrepancy / interplay between #129, #114 and #113...

lanceWilliams: @ludwigschwardt I think that the changes in [391](https://github.com/ska-sa/katcorelib/pull/391) can reasonably replace what was done in [113](https://github.com/ska-sa/astrokat/pull/113). This "slewonly" check would only need to ensure that it specifically passes track() a target dict with only an 'ants' key. Then there shouldn't be any impact on the rest of the observation setup or execution as far as I can tell.

lanceWilliams:
I think I am ok with approving https://github.com/ska-sa/katcorelib/pull/391, is this astrokat PR still being worked on/tested?

ludwigschwardt:
> I think I am ok with approving [ska-sa/katcorelib#391](https://github.com/ska-sa/katcorelib/pull/391), is this astrokat PR still being worked on/tested?

I think @tamerakassie could still address the minor suggestions you made.

astrokat/observe_main.py
ludwigschwardt: Should #129 then also effectively enable #114 on site, which hasn't been merged up to now? The `slewonly` functionality has never been used, in effect...

lanceWilliams: Yes, I think so. From distant memory I think this feature came out of trying to improve the startup for the PTUSE observation scripts in particular. I don't think other observations really experienced the same efficiency issue so there was no real pressure for it.

ludwigschwardt (approved):
Happy with this. @lanceWilliams, please just check that Tamera addressed your requested changes.

lanceWilliams (approved):
Looks ok

PR #133, opened 2026-05-05
Update SimSession AQF Tests Behavior to Match Katcorelib  MKAT Session 

nmakhoba:
This change updates how SimSession handles duration for reference pointing to match the katcorelib mkat session behavior.  

**Test: test_scans.py:TestAstrokatYAML.test_drift_scan_basic_sim**
- At the end of the observation the test looks for  `"Initiating 180-second track on target 'Az: -158:55:32.5 El: 52:01:18.0'"` but logs have `"initiating 180-second track on target 'Az: -159:01:46.8 El: 52:05:00.6'"`.
- The unittest simulator adds sleep time when updating observers time. The discrepancy in timing arises from how the simulator and Session dry run handle observer time. 

AQF pipeline progress output:

```
DRY-RUN: 2018-07-23 00:10:39.148Z INFO     Initialising Drift_scan target 1934-638 for 180.0 sec
DRY-RUN: 2018-07-23 00:10:39.148Z INFO     New compound scan: 'drift_scan'
DRY-RUN: 2018-07-23 00:10:39.148Z INFO     obs_start_ts 2018/7/23 00:10:00, transit_time:2018/7/23 00:11:30 az: 200:58:13.2 el: 52:05:00.6
DRY-RUN: 2018-07-23 00:10:39.148Z INFO     Drift_scan observation for 180.0 sec
DRY-RUN: 2018-07-23 00:10:39.148Z INFO     Initiating 180-second track on target 'Az: -159:01:46.8 El: 52:05:00.6' 
```

Unittest progress output:

```
2018-07-23 00:10:48Z - Initialising Drift_scan target 1934-638 for 180.0 sec
2018-07-23 00:12:18Z - obs_start_ts 2018/7/23 00:10:48, transit_time:2018/7/23 00:12:18 az: 201:04:27.5 el: 52:01:18.0 ref_ant:ref: 0-m dish at lat -30:42:39.8, long 21:26:38.0, alt 1086.6 m
2018-07-23 00:10:48Z - Drift_scan observation for 180.0 sec
2018-07-23 00:10:51Z - Slewed to Az: -158:55:32.5 El: 52:01:18.0 at azel (-158.9, 52.0) deg
```

- The Session dry run uses a fixed observer time of 00:10:00, however the simulator specifically adds a 48-second lead. 3 seconds during capture-init and 45 seconds for the target slew, this creates the time offset.

**Test: test_scans.py:TestAstrokatYAML.test_reference_pointing_scan_basic_sim**
- At the end of the observation the test looks for `'1934-638 observed for 144.0 sec'` in the logs but the logs have `'1934-638 observed for 288.0 sec'`  that's because the observation runs for 288 seconds instead of 144 seconds.
- SimSession treats the duration parameter as the time allocated for each track offset, the AQF interprets it as the total duration for the entire reference pointing scan. This mismatch causes the script to cycle through observations incorrectly.

More details about the SimSession and Mkat Session discrepancies are further explained [here](https://skaafrica.atlassian.net/browse/MT-7082)

### DIFF
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -337,7 +337,6 @@ class SimSession(object):
             len(offsets),
             track_duration,
         )
-        #self.track(target, duration=0, announce=False)
         for offset in offsets:
             user_logger.info("initiating track on offset of (%g, %g) degrees", *offset)
             self.track(target, track_duration, announce=False)
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -341,10 +341,6 @@ class SimSession(object):
             user_logger.info("initiating track on offset of (%g, %g) degrees", *offset)
             self.track(target, track_duration, announce=False)
         user_logger.info("Waiting for fitted beams to materialise in cal pipeline")
-        user_logger.info(
-            "reference time = %.1f, weather = %.1f deg C | %.1f hPa | %.1f %%",
-            # make these up...
-        )
         user_logger.info("Calculating and storing pointing offsets")
         return True
 
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -323,7 +323,7 @@ class SimSession(object):
         num_pointings: int
             Number of offset pointings
         """
-        scan = numpy.linspace(-extent, extent, (num_pointings -1) // 2)
+        scan = numpy.linspace(-extent, extent, (num_pointings - 1) // 2)
         offsets_along_x = numpy.c_[scan, numpy.zeros_like(scan)]
         offsets_along_y = numpy.c_[numpy.zeros_like(scan), scan]
         offsets = numpy.r_[offsets_along_y, offsets_along_x, [(0.0, 0.0)]]
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -50,7 +50,7 @@ def setobserver(update):
 
     """
     global simobserver
-    simobserver = update.copy()
+    simobserver = update
 
 
 def sim_time(record, datefmt=None):
@@ -375,9 +375,7 @@ class SimSession(object):
             ants_target = target.get('ants')
         else:
             ants_target = target
-        observer_date = target.antenna.observer.date
         az, el = ants_target.azel(simobserver.date)
-        target.antenna.observer.date = observer_date
         az = katpoint.rad2deg(az)
         el = katpoint.rad2deg(el)
         return az, el
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -50,7 +50,7 @@ def setobserver(update):
 
     """
     global simobserver
-    simobserver = update
+    simobserver = update.copy()
 
 
 def sim_time(record, datefmt=None):
@@ -323,26 +323,27 @@ class SimSession(object):
         num_pointings: int
             Number of offset pointings
         """
-        scan = numpy.linspace(-extent, extent, num_pointings // 2)
+        scan = numpy.linspace(-extent, extent, (num_pointings -1) // 2)
         offsets_along_x = numpy.c_[scan, numpy.zeros_like(scan)]
         offsets_along_y = numpy.c_[numpy.zeros_like(scan), scan]
-        offsets = numpy.r_[offsets_along_y, offsets_along_x]
+        offsets = numpy.r_[offsets_along_y, offsets_along_x, [(0.0, 0.0)]]
         offset_end_times = numpy.zeros(len(offsets))
         middle_time = 0.0
         weather = {}
+        track_duration = duration / num_pointings
 
         user_logger.info(
             "Initiating interferometric pointing scan on target "
             "'%s' (%d pointings of %g seconds each)",
             target.name,
             len(offsets),
-            duration,
+            track_duration,
         )
         self.track(target, duration=0, announce=False)
         # Point to the requested offsets and collect extra data at middle time
         for n, offset in enumerate(offsets):
             user_logger.info("initiating track on offset of (%g, %g) degrees", *offset)
-            self.track(target, duration, announce=False)
+            self.track(target, track_duration, announce=False)
             offset_end_times[n] = time.time()
             if n == len(offsets) // 2 - 1:
                 middle_time = offset_end_times[n]
@@ -374,7 +375,9 @@ class SimSession(object):
             ants_target = target.get('ants')
         else:
             ants_target = target
+        observer_date = target.antenna.observer.date
         az, el = ants_target.azel(simobserver.date)
+        target.antenna.observer.date = observer_date
         az = katpoint.rad2deg(az)
         el = katpoint.rad2deg(el)
         return az, el
--- a/astrokat/simulate.py
+++ b/astrokat/simulate.py
@@ -337,7 +337,7 @@ class SimSession(object):
             len(offsets),
             track_duration,
         )
-        self.track(target, duration=0, announce=False)
+        #self.track(target, duration=0, announce=False)
         for offset in offsets:
             user_logger.info("initiating track on offset of (%g, %g) degrees", *offset)
             self.track(target, track_duration, announce=False)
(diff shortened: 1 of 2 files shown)


astrokat/simulate.py
```diff
@@ -374,7 +375,9 @@ def _target_azel(self, target):
             ants_target = target.get('ants')
         else:
             ants_target = target
+        observer_date = target.antenna.observer.date
```
rishadebrahim: add a comment above this that mentions target.antenna.observer.date is destructively modified when calling target.azel.  This is similar to https://github.com/ska-sa/astrokat/blob/4c64bba93a1fd826c523306aee941d0c45b1c8d0/astrokat/scans.py#L25
The original target observer date is 'cached' and overwritten after the azel calculation.

Although I am not familiar with katpoint and astrokat, so I am not sure what the original intention of this code was and if the target observer date should be modified?

ludwigschwardt: I would not do that... This is neither the katpoint nor the astrokat API, but the pyephem way of doing things. The whole point of katpoint 😋 is to provide a unified API for basic coordinate operations. By talking to the underlying library like this, you just make the code harder to port to the next version of katpoint.

Let's find out why this was necessary and fix that spot instead...

ludwigschwardt (changes requested):

astrokat/simulate.py
rishadebrahim: @ludwigschwardt Thanks for commenting. One of the things i noted in my observation is that the az/el coordinates use the start_time in the yaml when using TimeSession vs SimSession which advances the target's observer date with every time.sleep for simulated delays for capture init and slewing before the coordinates are calculated.
The observer time for a target advances in SimSession, which does not happen on TimeSession.
This leads me to question which behaviour is the correct approach.

ludwigschwardt: I would not touch the target's observer date. It's internal state that should not be relied upon or used to transmit information. Can you point me to the code that updates it during time.sleep?

rishadebrahim: I think simobserver is being set to target observer's reference when https://github.com/ska-sa/astrokat/blob/4c64bba93a1fd826c523306aee941d0c45b1c8d0/astrokat/observatory.py#L280
is called.

So when time.sleep: https://github.com/ska-sa/astrokat/blob/4c64bba93a1fd826c523306aee941d0c45b1c8d0/astrokat/simulate.py#L183 is called anywhere in SimSession it advances the target observer date.

i added some logging in the method to compare the reference of simobserver and target.antenna.observer and they are the same object:

```python
    def _target_azel(self, target):
        """Get azimuth and elevation co-ordinates for a target at the current time.

        Parameters
        ----------
        target: katpoint.Target
            The target of interest.

        Returns
        -------
        az: float
            The azimuth co-ordinate of the target in degrees.
        el: float
            The elevation co-ordinate of the target in degrees.

        """
        match =  simobserver is target.antenna.observer
        user_logger.info("[DEBUG] Reference check for target observer vs simobserver: %s ", match)
        az, el = target.azel(simobserver.date)
        az = katpoint.rad2deg(az)
        el = katpoint.rad2deg(el)
        return az, el
```
Heres the logs without the changes added in this PR to simobserver and _target_azel methods:

```syslog
2018-07-23 00:10:00Z - Observation targets are ['1934-638']
2018-07-23 00:10:00Z - Waiting for observation setup
2018-07-23 00:10:03Z - INIT
2018-07-23 00:10:03Z - Slewing to first target
2018-07-23 00:10:03Z - [DEBUG] Reference check for target observer vs simobserver: True
2018-07-23 00:10:48Z - Slewed to 1934-638 at azel (200.8, 52.2) deg
2018-07-23 00:10:48Z - Tracked 1934-638 for 0 seconds
2018-07-23 00:10:48Z - Initialising Drift_scan target 1934-638 for 180.0 sec
2018-07-23 00:12:18Z - obs_start_ts 2018/7/23 00:10:48, transit_time:2018/7/23 00:12:18 az: 201:04:27.5 el: 52:01:18.0 ref_ant:ref: 0-m dish at lat -30:42:39.8, long 21:26:38.0, alt 1086.6 m
2018-07-23 00:10:48Z - Drift_scan observation for 180.0 sec
2018-07-23 00:10:48Z - [DEBUG] Reference check for target observer vs simobserver: True
2018-07-23 00:10:48Z - [DEBUG] Reference check for target observer vs simobserver: True
2018-07-23 00:10:51Z - Slewed to Az: -158:55:32.5 El: 52:01:18.0 at azel (-158.9, 52.0) deg
2018-07-23 00:13:51Z - Tracked Az: -158:55:32.5 El: 52:01:18.0 for 180 seconds
2018-07-23 00:13:51Z - Observation list completed - ending observation

```

astrokat/test/test_scans.py
```diff
@@ -32,7 +33,13 @@ def test_drift_scan_basic_sim(self):
         result = LoggedTelescope.user_logger_stream.getvalue()
         self.assertIn("Initialising Drift_scan target 1934-638 for 180.0 sec", result)
         self.assertIn("Drift_scan observation for 180.0 sec", result)
-        target_string = "Az: -158:55:32.5 El: 52:01:18.0"
+
+        # use running session to determine suitable target values
+        if os.getenv("SESSION_NAME") == "mkat_session":
```
rishadebrahim: A bit pedantic of me but, SESSION_TYPE is probably more appropriate.

Or checking for an environmental variable like AQF_RUN == True instead.

But this is fine.

rishadebrahim (approved):
Tested manually on devbox to confirm it works when not using AQF as the runner. Looks good to me

astrokat/test/test_scans.py
nmakhoba: @rishadebrahim this is a good suggestion, I'll make an update.

lanceWilliams (approved):
Looks ok

astrokat/simulate.py
```diff
@@ -323,26 +323,27 @@ def reference_pointing_scan(
...
             "Initiating interferometric pointing scan on target "
             "'%s' (%d pointings of %g seconds each)",
             target.name,
             len(offsets),
-            duration,
+            track_duration,
         )
         self.track(target, duration=0, announce=False)
         # Point to the requested offsets and collect extra data at middle time
         for n, offset in enumerate(offsets):
```
ludwigschwardt: No need for `n` or `offset_end_times` or `middle_time`.

astrokat/simulate.py
ludwigschwardt: This line is not in the session anymore. Also no extra `track()`. Have a look at `CaptureSession.reference_pointing_scan` for this part after the main for-loop too. I suggest:

```Python
        for offset in offsets:
            user_logger.info("initiating track on offset of (%g, %g) degrees", *offset)
            self.track(target, track_duration, announce=False)
        user_logger.info("Waiting for fitted beams to materialise in cal pipeline")
        user_logger.info(
            "reference time = %.1f, weather = %.1f deg C | %.1f hPa | %.1f %%",
            # make these up...
        )
        user_logger.info("Calculating and storing pointing offsets")
        return True
```

ludwigschwardt (changes requested):
Thanks! I would make `SimSession.reference_pointing_scan` still closer to `CaptureSession.reference_pointing_scan` to get better simulation results.

astrokat/simulate.py
```diff
@@ -323,35 +323,29 @@ def reference_pointing_scan(
...
-                    "reference time = %.1f, weather = %r", middle_time, weather)
-        user_logger.info("returning to target to complete the scan")
-        self.track(target, duration=0, announce=False)
-        user_logger.info("Waiting for gains to materialise in cal pipeline")
-        user_logger.info("Retrieving gains, fitting beams, storing offsets")
+            self.track(target, track_duration, announce=False)
+        user_logger.info("Waiting for fitted beams to materialise in cal pipeline")
+        user_logger.info(
+            "reference time = %.1f, weather = %.1f deg C | %.1f hPa | %.1f %%",
+            # make these up...
```
ludwigschwardt: You still have to make these up 😁

ludwigschwardt (changes requested):
One small change left.

ludwigschwardt (approved):
Nice, now you even have fewer lines of code than when you started: the best possible outcome. No code = no problems 😁

